PR #34, opened 2015-09-02
Video description cleansing/formatting like youtube-dl does

raltamirano:
Turning ```<br />```'s into newlines and unescaping character entities from the (HTML) description of videos.


package.json
```diff
@@ -6,7 +6,7 @@
     "video",
     "download"
   ],
-  "version": "0.5.1",
+  "version": "0.5.2",
```
fent: would rather update this myself as I use the command `npm version` that does it

lib/util.js
```diff
@@ -222,3 +222,22 @@ exports.parseFormats = function(info, debug) {
...
+ */
+exports.cleanVideoDescription = function(videoDescriptionHtml) {
+	if (!videoDescriptionHtml || videoDescriptionHtml == null) 
+		return null;
+
+	return entities.decode(videoDescriptionHtml
+		.replace('\n', ' ', 'gi')
+		.replace(/\s*\<\s*br\s*\/?\s*>\s*/gi, '\n')
+		.replace(/<\s*\/\s*p\s*>\s*<\s*p[^>]*>/gi, '\n')
+		.replace(/<.*?>/gi, '')
```
fent: why not use `<[^>]*>` like in the above regex?

lib/util.js
```diff
@@ -222,3 +222,22 @@ exports.parseFormats = function(info, debug) {
...
+/**
+ * @param {String} videoDescriptionHtml
+ * @return {String}
+ */
+exports.cleanVideoDescription = function(videoDescriptionHtml) {
+	if (!videoDescriptionHtml || videoDescriptionHtml == null) 
+		return null;
+
+	return entities.decode(videoDescriptionHtml
+		.replace('\n', ' ', 'gi')
```
fent: says [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace) flags in this function is deprecated.

test/util-test.js
```diff
@@ -229,3 +229,17 @@ describe('util.parseFormats()', function() {
...
+describe('util.cleanVideoDescription()', function() {
+  it('Clean the HTML for the video description', function() {
+    var data = require('./files/util/multiline-video-description.json');
+    var cleanDescription = util.cleanVideoDescription(data.html);
+    assert.ok(cleanDescription);
+    assert.equal(cleanDescription, "Some Title\n" +
+      "Line 1\n" +
+      "\"Line 2\"\n" +
+      "1  First Song  5:30\n" +
+      "2  Second Song  5:42");
```
fent: Can you use single quotes for this?

fent:
This is nice, thanks. Added some comments.

lib/util.js
raltamirano: Replaced with:

.replace('/\n/', ' ')

test/util-test.js
raltamirano: Sure.

lib/util.js
raltamirano: I can't fully understand these regexes, I've just blindly ported the idea from the "clean_html" in https://github.com/rg3/youtube-dl/blob/master/youtube_dl/utils.py - I'll need some help with this one.

lib/util.js
fent: `.replace(/\n/g, ' ')`

lib/util.js
raltamirano: After examining video descriptions lines like:

```2  Song name  <a href="#" onclick="yt.www.watch.player.seekTo(5*60+42);return false;">5:42</a><br>```

I think that the reasoning after those two differentes regexes is:

1. First, replace 'br' and an open + closing 'p' html tags with newlines (like the one at the end of the example line)
2. Remove ANY other html tag that remains after the previous step.

I think it works pretty well. Where do you exactly see ```<[^>]*>``` fit in those replacements? Let me know and I'll try to research some more on that direction.

File: lib/util.js [hunk 5]
```diff
@@ -222,3 +222,22 @@ exports.parseFormats = function(info, debug) {
...
+exports.cleanVideoDescription = function(videoDescriptionHtml) {
+	if (!videoDescriptionHtml || videoDescriptionHtml == null) 
+		return null;
+
+	return entities.decode(videoDescriptionHtml
+		.replace(/\n/g, ' ')
+		.replace(/\s*\<\s*br\s*\/?\s*>\s*/gi, '\n')
+		.replace(/<\s*\/\s*p\s*>\s*<\s*p[^>]*>/gi, '\n')
+		.replace(/<.*?>/gi, '')
+	).trim();
```
fent: Can you change these tabs to 2 spaces instead for consistency with the rest of the project?

raltamirano: Sure. Done!

fent:
Can you fix the conflicts? Also, there's another PR that uses the "entities" package instead, what do you think?

fent:
Thanks!

raltamirano:
Glad to help, thank *you* very much for this great tool! I'm using it as the guts for [ytdl-splitter-core] (https://github.com/raltamirano/ytdl-splitter-core), so I'll try my best to keep on collaborating as time allows.

PR #38, opened 2015-09-08
remove cheerio dependency

mallendeo:


lib/util.js
```diff
@@ -223,3 +224,18 @@ exports.parseFormats = function(info, debug) {
...
+
+/**
+ * Get video description from html
+ *
+ * @param {String} html
+ * @return {String}
+ */
+ exports.getVideoDescription = function(html) {
+   var regexString = '<p.*?id="eow-description".*?>(.+?)<\/p>'
+                   + '[\\n\\r\\s]*?<\/div>';
```
fent: why not make this a compile time regex object? rather than using the dynamic RegExp class.

lib/util.js
```diff
@@ -223,3 +224,18 @@ exports.parseFormats = function(info, debug) {
...
+   var regex = new RegExp(regexString, 'igm');
```
fent: I don't think the "g" option is needed here since it's only one match

lib/util.js
```diff
@@ -223,3 +224,18 @@ exports.parseFormats = function(info, debug) {
...
+   var description = regex.exec(html.match(regex));
```
fent: Using `RegExp#exec()` on the `String#match()` result??

package.json
```diff
@@ -17,7 +17,7 @@
     "test": "mocha -R spec -t 16000 test/*-test.js"
   },
   "dependencies": {
-    "cheerio": "~0.19.0",
+    "entities": "^1.1.1",
```
fent: There's currently another PR (#34) that uses html-entities instead, what do you think?

mallendeo: Updated. It'll be better for merge, also take less size in the browserify bundle, so it's fine.

Issue #43, opened 2015-10-04
Is there way to get the info of the video that error'ed?


opsigor:
Hi! I'm building an module to download the videos and convert them to mp3. I'm currently have an issue, because i'm downloading an entire playlist from youtube. 

Here is my code: 
```js
let streams = _vid.forEach((val) => {
    numVideos++;
    let _stream = ytdl(val, options);
    _stream
      .on('info', (info, format) => {
        fs.stat(path, (err, stats) => {
          if (err) {
            fs.mkdirSync(path);
          }
          self.emit('download', info.title);
          self.createSong(path, info.title, _stream, 192);
        });
      })
      .on('error', (err) => {
        self.emit('error', err);
        numVideos--;
      });
  })
``` 

This piece of code is inside of a forEach loop, and i was wondering if i could display the video that was errored to the user. For example, i'm brazilian and some videos are blocked on youtube on my country. When i try to download a video that is blocked, it throws:
``` 
[Error: Error 150: Este vídeo apresenta conteúdo de UMG, que o bloqueou no seu país com base nos direitos autorais.]
``` 
Is there i way that i could catch the name of the video?

Thanks!!!

fent:
The 'info' event is now emitted before the download is attempted.

fent:
Also, I'm wondering if there's a way around blocked videos. But I can't find one that is blocked in the US to test on.

PR #44, opened 2015-10-04
Added info on the error event

opsigor:
Referencing #43.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,8 +1,8 @@
 var PassThrough = require('stream').PassThrough;
-var request     = require('./request');
-var getInfo     = require('./info');
-var util        = require('./util');
-var cache       = require('./cache');
+var request = require('./request');
+var getInfo = require('./info');
+var util = require('./util');
+var cache = require('./cache');
 
 module.exports = ytdl;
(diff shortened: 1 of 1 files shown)


lib/index.js
```diff
@@ -1,8 +1,8 @@
 var PassThrough = require('stream').PassThrough;
-var request     = require('./request');
-var getInfo     = require('./info');
-var util        = require('./util');
-var cache       = require('./cache');
+var request = require('./request');
+var getInfo = require('./info');
+var util = require('./util');
+var cache = require('./cache');
```
fent: No need to remove all these spaces

lib/index.js
```diff
@@ -22,7 +22,10 @@ function ytdl(link, options) {
 
   getInfo(link, options, function(err, info) {
     if (err) {
-      stream.emit('error', err);
+      stream.emit('error', {
+        error: err,
+        info: info
+      });
```
fent: the "error" event should emit an error object. Does this work for you? Because I suspect that if there is an error here, then `info` won't be set to anything. Would be better if there "info" is emitted a few lines down, would be moved up and placed before `downloadFromInfoCallback()`

PR #68, opened 2016-02-12
will request more info if dashmpd isn't available initially

samelie:


fent:
The other URL, `/get_video_info`, is requested if the config isn't found

https://github.com/fent/node-ytdl-core/blob/master/lib/info.js#L70

So, in an effort to meet the programming rule of not repeating ourselves, I think it'd be best if this PR used the already existing code.

samelie:
The problem I was seeing was that even if the config exists and thus passes the option to request the `get_video_info`,  the dashmpd is no longer included by youtube in the `ytplayer.config`

fent:
Yeah I do think it is an issue. But I'm saying part of the PR already exists in the codebase, and should be used, instead of repeated.

samelie:
Okay let me adjust the code a bit, I  also just saw that vevo urls require a different youtube_info format

fent:
Yeah, the info format that's already in there is slightly different than this one here. I'm not sure if it matters too much, because the current one does work.

And.. the link to the youtube-dl description says that the dashmpd for get_info might be different. So, might be worth it to retrieve both of them and merge the formats.

samelie:
Just pushed a working fix. It doesn't do the mpd merging (didnt see your idea until after).
You thinking two separate fields?

Also the tests are borked, I think it's because like you said, the data is different.

lib/info.js
```diff
@@ -88,6 +87,48 @@ module.exports = function getInfo(link, opts, callback) {
...
+      callback();
+    }
+  }
+}
+
+/*
+youtube-dl 
+https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/youtube.py#L1127
+*/
+function getInfoUrl(id) {
```
fent: Since `requestAndMergeMpd()` is already a separated function, making another one for this that only gets called by it doesn't seem worth it.

lib/info.js
```diff
@@ -88,6 +87,48 @@ module.exports = function getInfo(link, opts, callback) {
...
+      });
+    } else {
+      callback();
+    }
+  }
+}
+
+/*
+youtube-dl 
+https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/youtube.py#L1127
```
fent: Would rather not link to another library to explain why something is done a certain way.

fent:
It looks like you're running your code through a formatter. It would be best to keep this file's style to fit the style of the rest of the files on the project.

As for the functionality, it might be worth to retrieve both dashmpd's from both sources, that is the video page and the info page. The async path logic for this can get tricky, since the embed page also requests the info page.

PR #74, opened 2016-06-01
Added support to retrieve the YouTube's suggested videos to the info object

paixaop:
Add the suggested videos to the info object. This allows users to automatically download related/suggested videos.

### DIFF
--- a/lib/info.js
+++ b/lib/info.js
@@ -37,6 +37,7 @@ module.exports = function getInfo(link, options, callback) {
 
   var myrequest = options.request || request;
   var id = util.getVideoID(link);
+  var relatedIds = [];
   util.parallel([
     function(callback) {
       // Try getting config from the video page first.
@@ -46,7 +47,20 @@ module.exports = function getInfo(link, options, callback) {
 
         // Get description from #eow-description
         var description = util.getVideoDescription(body);
+        var related = '';
+        var ids = '';
+        if( body.search(/RELATED_PLAYER_ARGS/) ) {
+          related = body.match(/RELATED_PLAYER_ARGS['\s\t\n]*:[\s\t\n]*{[\s\t\n"]*rvs":"([^}]+)/)[1];
 
+          var re= /id=([a-zA-Z0-9_-]+)/gm;
+          var m;
+          while ((m = re.exec(related)) !== null) {
+              if (m.index === re.lastIndex) {
+                  re.lastIndex++;
+              }
+              relatedIds.push(m[1]);
+          }
+        }
         var jsonStr = util.between(body, 'ytplayer.config = ', '</script>');
         if (jsonStr) {
           jsonStr = jsonStr.slice(0, jsonStr.lastIndexOf(';ytplayer.load'));
@@ -108,6 +122,7 @@ module.exports = function getInfo(link, options, callback) {
     } else {
       dashmpd2 = results[1].dashmpd;
     }
+    config.args.relatedIds = relatedIds;
     gotConfig(options, description, config, dashmpd2, callback);
   });
 };
--- a/lib/info.js
+++ b/lib/info.js
@@ -37,7 +37,7 @@ module.exports = function getInfo(link, options, callback) {
 
   var myrequest = options.request || request;
   var id = util.getVideoID(link);
-  var relatedIds = [];
+  var relatedVideos = [];
   util.parallel([
     function(callback) {
       // Try getting config from the video page first.
@@ -47,7 +47,7 @@ module.exports = function getInfo(link, options, callback) {
 
         // Get description from #eow-description
         var description = util.getVideoDescription(body);
-        relatedIds = getRelatedVideos(body);
+        relatedVideos = getRelatedVideos(body);
 
         var jsonStr = util.between(body, 'ytplayer.config = ', '</script>');
         if (jsonStr) {
@@ -69,7 +69,7 @@ module.exports = function getInfo(link, options, callback) {
           url = EMBED_URL + id;
           myrequest(url, options.requestOptions, function(err, body) {
             if (err) return callback(err);
-
+            relatedVideos = getRelatedVideos(body);
             var html5player = util.between(body, '"js":"', '"');
             if (!html5player) {
               return callback(new Error('Could not find `player config`'));
@@ -96,7 +96,6 @@ module.exports = function getInfo(link, options, callback) {
       });
       myrequest(url, options.requestOptions, function(err, body) {
         if (err) return callback(err);
-        relatedIds = getRelatedVideos(body);
         callback(null, querystring.parse(body));
       });
     }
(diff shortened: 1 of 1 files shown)


lib/info.js
```diff
@@ -46,7 +47,20 @@ module.exports = function getInfo(link, options, callback) {
 
         // Get description from #eow-description
         var description = util.getVideoDescription(body);
-
+        var related = '';
+        var ids = '';
+        if( body.search(/RELATED_PLAYER_ARGS/) ) {
+          related = body.match(/RELATED_PLAYER_ARGS['\s\t\n]*:[\s\t\n]*{[\s\t\n"]*rvs":"([^}]+)/)[1];
```
fent: Tests fail from this line, make sure to check if there is a match.

I'm also not a big fan of this regexp, for one, singlequotes always surround `RELATED_PLAYER_ARGS` right? And `\s` also matches `\t` and `\n`. `rvs` is also always surrounde by doublequotes, it looks like.

lib/info.js
```diff
@@ -37,6 +37,7 @@ module.exports = function getInfo(link, options, callback) {
 
   var myrequest = options.request || request;
   var id = util.getVideoID(link);
+  var relatedIds = [];
```
fent: I don't agree with putting this here, it feels out of the way. Would make more sense if it came before the loop.

lib/info.js
```diff
@@ -46,7 +47,20 @@ module.exports = function getInfo(link, options, callback) {
 
         // Get description from #eow-description
         var description = util.getVideoDescription(body);
-
+        var related = '';
```
fent: Doesn't need to be declared this early.

lib/info.js
```diff
@@ -46,7 +47,20 @@ module.exports = function getInfo(link, options, callback) {
...
+        var ids = '';
```
fent: This isn't used?

lib/info.js
```diff
@@ -46,7 +47,20 @@ module.exports = function getInfo(link, options, callback) {
...
+        if( body.search(/RELATED_PLAYER_ARGS/) ) {
+          related = body.match(/RELATED_PLAYER_ARGS['\s\t\n]*:[\s\t\n]*{[\s\t\n"]*rvs":"([^}]+)/)[1];
+
+          var re= /id=([a-zA-Z0-9_-]+)/gm;
+          var m;
+          while ((m = re.exec(related)) !== null) {
+              if (m.index === re.lastIndex) {
```
fent: I'm not sure why this is needed?

fent:
Thanks for taking the effort to create the PR.

I'm hesitant about this one, mainly because there's possibility that the video is blocked to non-logged in users, and the embed page would be used to get the video config. In that case, related video IDs would not be retrievable, unless that exists in any of the other endpoints?

I'm also not too into getting this through regex, but it's not a big deal.

Your coding style doesn't match the rest of the project's style. It's a good idea to keep the entire project's codebase consistent, making it look as if it was created by one person alone. Make sure to keep in mind things like spaces around parentheses, before equal signs, indent width.

paixaop:
I have updated and cleaned up the code. All tests pass now. 
tried to match the coding style let me know if you see anything out of place.

I tested the code logged in and logged out to see if it worked and it seems to be OK.

When I test with the blocked video 'rIqCiJKWx9I' it did not find any related items as the embedded player only shows them at the end of the playback. Not sure how we could get them without

lib/info.js
```diff
@@ -108,10 +111,32 @@ module.exports = function getInfo(link, options, callback) {
...
     gotConfig(options, description, config, dashmpd2, callback);
   });
 };
 
+/**
+ * @param {String} HTML body
+ */
+function getRelatedVideos(body) {
+  // Get the RELATED_PLAYER_ARGS from body then parse for the video Ids
+  var related = body.match(/RELATED_PLAYER_ARGS['\s\t\n]*:[\s\t\n]*{[\s\t\n"]*rvs":"([^}]+)/);
```
fent: Still think this regex can be simplified. And since the format of the "rvs" string seems consistent, we could probably use something like `related[1].split(',').map(function(s) { return querystring.parse(s); })` which would also retrieve a bit for data of the related videos.

lib/info.js
```diff
@@ -94,6 +96,7 @@ module.exports = function getInfo(link, options, callback) {
       });
       myrequest(url, options.requestOptions, function(err, body) {
         if (err) return callback(err);
+        relatedIds = getRelatedVideos(body);
```
fent: The body of this request is a querystring, no need to look for related videos here.

lib/info.js
```diff
@@ -108,10 +111,32 @@ module.exports = function getInfo(link, options, callback) {
     } else {
       dashmpd2 = results[1].dashmpd;
     }
+    config.args.relatedIds = relatedIds;
```
fent: Would be cleaner to pass this along in the same way `description` is.

lib/info.js
```diff
@@ -108,10 +111,32 @@ module.exports = function getInfo(link, options, callback) {
...
     gotConfig(options, description, config, dashmpd2, callback);
   });
 };
 
+/**
+ * @param {String} HTML body
```
fent: Just "body" here, not "HTML body".

Description of a parameter comes after the parameter name.

paixaop:
Removed the regex code completely to parse the related player and used `querystring.parse` on `rvs` as you suggested, but it turns out the the string is not so well formatted and sometimes youtube inserts a `,` without using `&` to separate query string fields. 

Example: 

    ...6autonav%3D1%26playnext%3D1,author=World`

So sometimes the `querystring.parse()` will not return the correct objects. What do you suggest we do to fix this issue? Any ideas?

fent:
My comment about the regex also suggested doing `.split(',')` on the string.

`description` is passed to the callback. The callback only takes one "results" argument, so I made it an array to keep it simple. Then later on, `description` is retrieved from `results[0][0]`, and then assigned to `info.description`.

Maybe I should make that part better by allowing any number of params to the callback.

paixaop:
Too tired last night, added the split(',') this AM and it works great

lib/info.js
```diff
@@ -37,6 +37,7 @@ module.exports = function getInfo(link, options, callback) {
 
   var myrequest = options.request || request;
   var id = util.getVideoID(link);
+  var relatedVideos = null;
```
fent: Still don't think this should be placed here, as it only makes sense where it's retrieved from.

lib/info.js
```diff
@@ -108,10 +114,41 @@ module.exports = function getInfo(link, options, callback) {
     } else {
       dashmpd2 = results[1].dashmpd;
     }
-    gotConfig(options, description, config, dashmpd2, callback);
+    gotConfig(options, description, relatedVideos, config, dashmpd2, callback);
   });
 };
 
+/**
+ * @param {String} body
```
fent: Forgot to comment about this earlier, this should also have

```* @return {Array.<Object>}```

File: lib/info.js [hunk 12]
```diff
@@ -108,10 +114,41 @@ module.exports = function getInfo(link, options, callback) {
...
+ */
+function getRelatedVideos(body) {
+  var jsonStr = util.between(body, 'yt.setConfig(', '</script>');
+  var related = [];
+  if (jsonStr) {
+    try {
+      jsonStr = jsonStr.slice(0, jsonStr.indexOf(');'));
```
fent: If you're doing this here, why not put the string ");" in the above call to `util.between()`

lib/info.js
```diff
@@ -108,10 +114,41 @@ module.exports = function getInfo(link, options, callback) {
...
+      jsonStr = jsonStr.replace(/'/g,'"');
+      var s = jsonStr.split(',');
+      var config = JSON.parse(jsonStr);
+      if (config && config.RELATED_PLAYER_ARGS && config.RELATED_PLAYER_ARGS.rvs) {
+        var qStrs = config.RELATED_PLAYER_ARGS.rvs.split(',');
+        for (i in qStrs) {
```
fent: `rvs` is an array, it shouldn't be looped using for in syntax, that's for objects only. You can use use `Array#map()`

lib/info.js
```diff
@@ -108,10 +114,41 @@ module.exports = function getInfo(link, options, callback) {
...
+          related.push({
+            id: q.id,
+            author: q.author,
+            title: q.title,
+            length_seconds: q.length_seconds
+          });
+        }
+      }
+    }
+    catch (err) {
```
fent: Put the try/catch statement around the `JSON.parse()` call, since that's the only line that would be throwing an error.

fent:
Made a few more comments. Sorry for not getting to this earlier.

frash23:
Is this still planned to be implemented? I could really use it.
If this patch has been orphaned I am interested in adopting it.

paixaop:
@frash23 if you want to make the final adjustments please do so as I don't have the time right now. As is the implementation works and that's what I used in my project, but I don't have the time to get the patch ready for merging.

fent:
Sorry for not being able to accept this PR as it is. I can imagine how frustrating it can be to go as far as commit something you worked on, on someone else's project, with the intention on sharing your contributions with that person and with everyone that uses that library. Only to have your contribution be criticized.

And thanks for going that far. Most won't submit a PR, they only create an issue, which is still better than nothing (if it is an issue).

But, I cannot accept the changes as they are. A project's code should look like it was developed by one unified group, so that when someone looks through its code, they can focus on what the program is doing rather than notice the differences between dev A and dev B. These style changes are not just spacing, and brackets vs no-brackets, but also best practices, and even how logic is organized.

If you don't have time to make the changes, it's cool.

PR #76, opened 2016-06-03
Change tag names to uppercase in getDashManifest

depr:
When sax.parser is not in strict mode all attributes and tag names will be in uppercase.


depr:
There you go. Now, since getDashManifest started to work it found one new format (itag 141) that hadn't been in the test results. So I had to change them too.

fent:
Thank you!

PR #77, opened 2016-06-05
added options parameters

EdsonAlcala:


fent:
Good catch, thank you.

PR #78, opened 2016-06-14
Added Typescript definitions.

JimmyBoh:
I have started working with Typescript and it really has made development a lot easier/more efficient thanks to excellent type guidance and other error checking. I noticed that `node-ytdl-core` didn't have any definition files, so I've created the necessary changes to enable anyone using TypeScript to get proper intellisense. Please review the names for everything, some of the interfaces could probably get a better name/description. Also if there are key properties missing I can add them.

Once merged, the only thing that must be done is to [register this module in the Typings Registry](https://github.com/typings/registry#contribute-yourself). I would be happy to set that up, but I figured you'd like to review (and I revise) this PR first.

### DIFF
--- a/_references.d.ts
+++ b/_references.d.ts
@@ -1,0 +1,1 @@
+/// <reference path="./typings/index.d.ts" />
--- a/typings.json
+++ b/typings.json
@@ -1,0 +1,8 @@
+{
+  "name": "ytdl-core",
+  "main": "./lib/index.d.ts",
+  "dependencies": {},
+  "globalDependencies": {
+    "node": "registry:dt/node#6.0.0+20160613154055"
+  }
+}
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 node_modules
+typings
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -1,0 +1,113 @@
+import {ClientRequest} from 'http';
+import {Readable} from 'stream';
+
+/**
+ * Attempts to download a video from the given url.
+ */  
+export function ytdl(link: string, options?: DownloadOptions): Readable;
+export namespace ytdl
+{
+  /**
+   * Use this if you only want to get metainfo from a video.
+   */  
+  export function getInfo(url: string, callback: (err: any, info: YoutubeInfo) => void);
+
+  /**
+   * Use this if you only want to get metainfo from a video.
+   */
+  export function getInfo(url: string, options: DownloadOptions, callback: (err: any, info: YoutubeInfo) => void);
+
+  /**
+   * Once you have received metadata from a video with the `getInfo` function, you may pass that `info`, along with other `options` to `downloadFromInfo`.
+   */  
+  export function downloadFromInfo(info: YoutubeInfo, options?: DownloadOptions): Readable;
+}
+
+export interface YoutubeFormat {
+  fallback_host: string;
+  quality: string;
+  type: string;
+  url: string;
+  itag: string;
+  sig: string;
+  container: 'flv' | '3gp' | 'mp4' | 'webm' | 'ts' | string;
+  resolution: '144p' | '240p' | '270p' | '360p' | '480p' | '720p' | '1080p' | '1440p' | '2160p' | '3072p' | string;
+  encoding: 'VP8' | 'vp8' | 'VP9' | 'H.264' | 'MPEG-4 Visual' | 'Sorenson H.283' | 'Sorenson H.263' | string;
+  profile: '3d' | 'high' | 'main' | 'simple' | 'baseline' | 'Main@L3.1' | string;
+  bitrate: string;
+  audioEncoding: 'mp3' | 'aac' | 'opus' | 'vorbis' | string;
+  audioBitrate: number;
+}
+
+export interface YoutubeInfo {
+  plid: string;
+  sendtmp: number;
+  title: string;
+  status: string;
+  muted: number;
+  pltype: string;
+  thumbnail_url: string;
+  has_cc: boolean;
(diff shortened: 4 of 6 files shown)


fent:
I've used typescript before, but haven't made definition files along with it. But I'm confused, that registry repo says otherwise on how to submit a typing. Do the definitions get submitted to that repo, or to a separate new repo?

Why is the "typings" folder ignored in `.gitignore`? Why is there a "typings" field in `package.json`? Why include the typings package in the dependencies?

AS for the interface names, I'd only change `YoutubeInfo` and `YoutubeFormat` to `VideoInfo` and `VideoFormat`.

JimmyBoh:
> But I'm confused, that registry repo says otherwise on how to submit a typing. Do the definitions get submitted to that repo, or to a separate new repo?

They are a little confusing, but it's neither actually. They stay packaged with your library ([read the first option in the "How?" section of this issue](https://github.com/typings/typings/issues/322))

> Why is the "typings" folder ignored in .gitignore?

`typings` (library) is to external TypeScript definitions as `npm` is to external node modules, they both manage dependencies of the current project. The same reason `node_modules` is excluded from git applies to `typings` (folder) as well (unless you want to check in your dependencies folders).

> Why is there a "typings" field in package.json?

Related to the answer above, this tells the TypeScript compiler where to locate the definitions for this library (whether using `typings` (library) or not). [See this brief but informative article for more details](https://www.typescriptlang.org/docs/handbook/typings-for-npm-packages.html).

> Why include the `typings` package in the dependencies?

As mentioned before, `typings` manages external typescript definitions. With this project, it just needs `node`, but other projects might have significantly more. By including `typings` in the `devDependencies` any contributor can easily maintain the definitions of this project. One of the things I forgot to add was a `postinstall` script that runs `typings install`(to install any missing dependencies). As it stands now, if you don't run `typings install` then the typescript definition file will give errors (it doesn't know it's in node land). This is only a problem if you try to update the definition file, it has zero effect on the library itself.

I will add the correct `postinstall` step and change the names of the interfaces. I will let you know when it is ready.

fent:
I took a lot through that typings issue to get a sense of how this has been done in other libraries. Judging from the other PRs to other libs, it looks like this PR has a few too many files. [This PR to moment](https://github.com/moment/moment/pull/2916) for example, only includes the typings definition file, and a reference to it in `package.json`.

From https://github.com/typings/typings/issues/322
> There's two way we can collaborate with package authors. The preferred is that we can help get type definition inside packages. This means consumers don't even need to install Typings, and things start working immediately after they install a package. 

So it looks like including typings in devDependencies is not required.

JimmyBoh:
I believe the PR is correct, but @typings please step in if I have made any mistakes. 

`moment` is different than `node-ytdl-core` in that `moment` has zero dependencies. So when developing the typescript definitions for `moment` there is no need to use `typings` to pull in definitions of dependencies because there are none. `node-ytdl-core` does have dependencies (`node`, specifically `stream.Readable` and `http.ClientRequest` for the external methods only). 

Without `typings` the following imports would just be an error since the TypeScript compiler has nothing to tell it what `http` and `stream` are:

```js
import {ClientRequest} from 'http';
import {Readable} from 'stream';
```

Once you use `typings` to install `node` as a definition dependency the TypeScript compiler will have a reference to both of the libraries (and more) and the import will resolve correctly.

As a test, clone my PR and (without calling `npm install` or `typings install`) open the project in an editor that supports TypeScript (I highly recommend Visual Studio Code). It should show both of the import statements above as "Can't find module 'http'" or something along the lines of that. Then call `npm install` (which should automatically call `typings install`) and wait for the definition files to be imported. Once they are, those errors should be resolved, and you are provided typings for all standard node modules (`fs`, `path`, etc).

typings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "ytdl-core",
+  "main": "./lib/index.d.ts",
+  "dependencies": {},
+  "globalDependencies": {
+    "node": "registry:dt/node#6.0.0+20160613154055"
```
fent: Is this required, this module should work with node versions specified in `.travis-ci`. With typings, would it only work in nodev6? And why point to a release this specifically?

_references.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference path="./typings/index.d.ts" />
```
fent: Wondering if this file is also required. I didn't see it in other typings PRs.

JimmyBoh: This one I am less confident about, but I know a lot of editors will use the `_references.d.ts` file as the source of all definition files. This just specifies the main `typings` file which then loads each of the downloaded dependencies.

typings.json
JimmyBoh: You have to remember that `typings` and the `typings.json` is only used when a developer is **contributing** to `node-ytdl-core`, not when it is used within another node app (or when unit tests are ran on Travis CI). The version _could_ be rolled back to `4.0.0`, but that really doesn't do much, since the utilized API for `http.ClientRequest` and `stream.Readable` hasn't changed between `4.0.0` and `6.0.0`.

It is very difficult for me to explain how to develop tools to make it easier to develop tools, but [maybe this explanation will help](https://github.com/typings/typings/blob/master/docs/faq.md#why). Maybe the best idea is to just pull it down, and mess around (delete some files, change some code, removed dependencies, etc). That will really give you a good idea as to what each files does.

JimmyBoh:
So TypeScript is now at 2.0, which includes changes as to how definition files are exported (it's much simpler). I will take a look at how different it really is and open a new PR to handle it instead.

fent:
Sounds good

fent:
Closing this due to merging #137

PR #85, opened 2016-10-22
Match raw video ids

Guichaguri:
Currently, ytdl-core only accepts URLs, this PR allows it to accept raw IDs such as `ysz0RxGyR_Y` using a fast regular expression.


fent:
Thanks! Merged and published.

PR #92, opened 2016-12-22
Support web streaming to html5 video tag 

sub2u:
In my requirement i am streaming the response to html5 video tag.

I will update the documentation for we streaming code, if you agree with my patch.

Thanks,


fent:
Thank you! New version published.

PR #101, opened 2017-01-26
Added support for date-published, owner and related-videos

TimeForANinja:
credits for related videos go to [paixaop](https://github.com/paixaop)


package.json
```diff
@@ -6,7 +6,7 @@
     "video",
     "download"
   ],
-  "version": "0.7.22",
+  "version": "0.7.23",
```
fent: Don't include the new version number in the diff. This is usually done from the command `npm version`

lib/util.js
```diff
@@ -291,6 +291,83 @@ exports.getVideoDescription = function(html) {
...
+};
+
+
+/**
+ * Get video published at from html
+ *
+ * @param {String} body
+ * @return {Array.<Object>}
+ */
+//credits to https://github.com/paixaop
```
fent: This is better suited inside the jsdoc description above.

lib/util.js
```diff
@@ -291,6 +291,83 @@ exports.getVideoDescription = function(html) {
...
+exports.getRelatedVideos = function(body) {
+  jsonStr = this.between(body, 'yt.setConfig(', ');')
```
fent: If we're only interested in `RELATED_PLAYER_ARGS.rvs`, let's start our search there, from there til `"},` seems like it would work.

Add a semicolon at the end of the line.

lib/util.js
```diff
@@ -291,6 +291,83 @@ exports.getVideoDescription = function(html) {
...
+    var s = jsonStr.split(',');
+    try {
+      var config = JSON.parse(jsonStr);
+    }
+    catch (err) {
+      return [];
+    }
+    if (config && config.RELATED_PLAYER_ARGS && config.RELATED_PLAYER_ARGS.rvs) {
+      var qStrs = config.RELATED_PLAYER_ARGS.rvs.split(',');
+      if(qStrs.length > 0) {
```
fent: No need to check for length here, `Array#map` will return empty if the list is empty anyway.

lib/util.js
```diff
@@ -291,6 +291,83 @@ exports.getVideoDescription = function(html) {
...
+        related = qStrs.map(function(relatedVideo){
+		var q = qs.parse(relatedVideo)
+		return {
+          'id':q.id,
+          'author':q.author,
+          'title':q.title,
+          'length_seconds':q.length_seconds
```
fent: We can return the full parsed object.

lib/info.js
```diff
@@ -144,7 +153,9 @@ function gotConfig(id, options, description, config, callback) {
     }
 
     info.formats = util.parseFormats(info);
-    info.description = description;
+
+    //add the additional to the infos and keep it future proof
+    info = Object.assign({}, info, additional);
```
fent: `Object.assign` is not in node v0.12, it's why the tests are failing for that version.

I'd argue that it might be time to stop supporting such an old version. But this functionality is also very simple to replicate with a for loop.

TimeForANinja:
First off - thanks for your feedback
yes it is my first PR and i'm happy to get serious feedback

tryed my best to match ur styling
util.getRelatedVideos() was shortent thanks to ur comments
util.getAuthor() got a more or less rewrite
swapped out the Object.assign() to support v0.12 again

lib/util.js
```diff
@@ -291,6 +309,65 @@ exports.getVideoDescription = function(html) {
...
+exports.getRelatedVideos = function(body) {
+  var jsonStr = this.between(body, '\'RELATED_PLAYER_ARGS\': {"rvs":','},')
+  try {
+    jsonStr = JSON.parse(jsonStr)
+  }
+  catch(err) {
+	return [];
+  }
+  var unparsed_videos = jsonStr.split(',')
+  return unparsed_videos.map(function(link) {
```
fent: Since it the `unparsed_videos` variable will only be used once and in the line right after, you can put `jsonStr.split(',')` here instead.

lib/util.js
```diff
@@ -291,6 +309,65 @@ exports.getVideoDescription = function(html) {
...
+ */
+var authorRegexp = /<a href="\/channel\/([\w-]+)"[^>]+>(.+?(?=<\/a>))/
+var aliasRegExp = /<a href="\/user\/([^"]+)/
+exports.getAuthor = function(body) {
+  var ownerinfo = this.between(body, '<div id="watch7-user-header" class=" spf-link ">', '<div id="watch8-action-buttons" class="watch-action-buttons clearfix">');
+  ownerinfo = new Entities().decode(ownerinfo)
+  if (ownerinfo === '') {
+	  return {};
+  }
+  let match = ownerinfo.match(authorRegexp);
```
fent: `let` isn't supported in node v0.12

lib/info.js
```diff
@@ -50,9 +50,22 @@ module.exports = function getInfo(link, options, callback) {
         return callback(new Error('Video not found'));
       }
     }
-
-    // Get description from #eow-description.
-    var description = util.getVideoDescription(body);
+	
+	
+
+
```
fent: Why so many new lines?

lib/util.js
```diff
@@ -233,6 +248,9 @@ exports.getVideoID = function(link) {
   if (!id) {
     throw new Error('No video id found: ' + link);
   }
+  if(id === 'playlist') {
+    throw new Error('Playlists not supported (yet)');
+  }
```
TimeForANinja: maybe thatll come one day... should fix the error descriped in https://github.com/fent/node-ytdl-core/issues/102

lib/util.js
```diff
@@ -291,6 +309,65 @@ exports.getVideoDescription = function(html) {
...
+var aliasRegExp = /<a href="\/user\/([^"]+)/
+exports.getAuthor = function(body) {
+  var ownerinfo = this.between(body, '<div id="watch7-user-header" class=" spf-link ">', '<div id="watch8-action-buttons" class="watch-action-buttons clearfix">');
+  ownerinfo = new Entities().decode(ownerinfo)
+  if (ownerinfo === '') {
+	  return {};
+  }
+  let match = ownerinfo.match(authorRegexp);
+  return {
+	'type':'channel',
```
TimeForANinja: Kinda... the channel obj seems given for all channels and youtube uses it for the "uploads of" playlist. The user is  optional so i put it in the aka

lib/util.js
fent: Even so, a fix for that should be isolated to its own PR.

lib/util.js
fent: I meant, it seems superfluous to put it there since it will always be of type 'channel'.

lib/util.js
```diff
@@ -291,6 +306,63 @@ exports.getVideoDescription = function(html) {
...
+ * @return {Object}
+ */
+var authorRegexp = /<a href="\/channel\/([\w-]+)"[^>]+>(.+?(?=<\/a>))/
+var aliasRegExp = /<a href="\/user\/([^"]+)/
+exports.getAuthor = function(body) {
+  var ownerinfo = this.between(body, '<div id="watch7-user-header" class=" spf-link ">', '<div id="watch8-action-buttons" class="watch-action-buttons clearfix">');
+  ownerinfo = new Entities().decode(ownerinfo)
+  if (ownerinfo === '') {
+    return {};
+  }
```
fent: You can move this if statement one line above, since decoding entities won't matter if nothing was found by `between`

lib/util.js
```diff
@@ -291,6 +306,63 @@ exports.getVideoDescription = function(html) {
...
+ * @param {String} body
+ * @return {Array.<Object>}
+ */
+exports.getRelatedVideos = function(body) {
+  var jsonStr = this.between(body, '\'RELATED_PLAYER_ARGS\': {"rvs":', '},')
+  try {
+    jsonStr = JSON.parse(jsonStr)
+  }
+  catch (err) {
+	return [];
```
fent: Indentation is inconsistent here

lib/util.js
```diff
@@ -291,6 +306,63 @@ exports.getVideoDescription = function(html) {
...
 /**
+ * Get video Owner from html.
+ *
+ * @param {String} body
+ * @return {Object}
+ */
+var authorRegexp = /<a href="\/channel\/([\w-]+)"[^>]+>(.+?(?=<\/a>))/
+var aliasRegExp = /<a href="\/user\/([^"]+)/
+exports.getAuthor = function(body) {
+  var ownerinfo = this.between(body, '<div id="watch7-user-header" class=" spf-link ">', '<div id="watch8-action-buttons" class="watch-action-buttons clearfix">');
```
fent: I didn't notice before, but you should refer to the `between` function by `exports.between` so that it's more explicit, and just in case someone ever calls `getAuthor` with a different context.

lib/util.js
```diff
@@ -291,6 +306,63 @@ exports.getVideoDescription = function(html) {
...
+  if (ownerinfo === '') {
+    return {};
+  }
+  var match = ownerinfo.match(authorRegexp);
+  return {
+    ref: '/channel/'+match[1],
+    id: match[1],
+    name: match[2],
+    avatar: this.between(ownerinfo, 'data-thumb="', '"'),
+    user: (ownerinfo.match('/user/') ? '/user/'+ownerinfo.match(aliasRegExp)[1] : 'none'),
```
fent: Spaces around `+`

The parenthesis around this ternary statement are not needed here, and it looks it's as clear without it too.

Should use `null` so that no user doesn't point to https://www.youtube.com/user/none. If given `null`, someone can more easily check for that. Also, how often does the user not get matched? Maybe a test could be added where it finds the channel but no user?

test/util-test.js
```diff
@@ -219,6 +219,21 @@ describe('util.between()', function() {
 });
 
 
+describe('util.objectAssign()', function() {
+  if('Merges to objects into one', function() {
```
fent: `if`?

test/util-test.js
```diff
@@ -219,6 +219,21 @@ describe('util.between()', function() {
...
+    var obj1 = {key1:'value1', key2:true}
+    var obj2 = {key2:false, key3:[1,2,{3:3}], key4:'value4'}
+    var mergedObject = util.objectAssign(obj1,obj2)
+    assert.deepEqual(mergedObject, {
+      key1:'value1',
+      key2:false,
+      key3:[1,2,{3:3}],
+      key4:'value4'
```
fent: Spacing around these object declarations doesn't match the style for the rest of the project. The project uses spaces after `{` and before `}` for objects on the same line. And spaces after `:` and `,`, except at the end of the line.

test/util-test.js
```diff
@@ -261,6 +276,65 @@ describe('util.getVideoDescription()', function() {
...
+      'files/util/related-video'), 'utf8');
+    var authorObj = util.getAuthor(html);
+    assert.deepEqual(authorObj,{
+      type: 'channel',
+      ref: '/channel/UC_aEa8K-EOJ3D6gOs7HcyNg',
+      id: 'UC_aEa8K-EOJ3D6gOs7HcyNg',
+      name: 'NoCopyrightSounds',
+      avatar: 'hisprofile.pic',
+      user: '/user/NoCopyrightSounds',
+	});
```
fent: Indenting here inconsistent

test/util-test.js
```diff
@@ -261,6 +276,65 @@ describe('util.getVideoDescription()', function() {
 });
 
 
+describe('util.getAuthor()', function() {
+  it('Retrieves formatted video author', function() {
+    var html = fs.readFileSync(path.resolve(__dirname,
+      'files/util/related-video'), 'utf8');
+    var authorObj = util.getAuthor(html);
+    assert.deepEqual(authorObj,{
```
fent: Space after `,`

test/util-test.js
```diff
@@ -261,6 +276,65 @@ describe('util.getVideoDescription()', function() {
...
+  });
+})
+
+
+describe('util.getPublished()', function() {
+  it('Retrieves formatted published date', function() {
+    var html = fs.readFileSync(path.resolve(__dirname,
+      'files/util/related-video'), 'utf8');
+    var publishedTimestamp = util.getPublished(html);
+    assert.equal(publishedTimestamp.toString(),'1416355200000');
```
fent: Space after `,`

You can compare the ints directly without converting to string.

test/util-test.js
```diff
@@ -261,6 +276,65 @@ describe('util.getVideoDescription()', function() {
...
+  });
+})
+
+
+describe('util.getRelatedVideos()', function() {
+  it('Retrieves formatted video author', function() {
+    var html = fs.readFileSync(path.resolve(__dirname,
+      'files/util/related-video'), 'utf8');
+    var authorObj = util.getRelatedVideos(html);
```
fent: `authorObj` should be something more related, like `relatedVideos`, or you can inline it in `assert.deepEqual`

test/util-test.js
```diff
@@ -261,6 +276,65 @@ describe('util.getVideoDescription()', function() {
...
+      },
+      {
+        playlist_title: "Mix – Alan Walker - Fade [NCS Release]",
+        list: "RDbM7SZ5SBzyY",
+        playlist_iurlmq: "iurlmq2",
+        session_data: "itct=firstvid%3D%3D",
+        playlist_length: "0",
+        thumbnail_ids: "AOeY-nDp7hI",
+        video_id: "AOeY-nDp7hI",
+        playlist_iurlhq: "second.pic",
```
fent: Use single quotes on these

lib/util.js
```diff
@@ -291,6 +306,63 @@ exports.getVideoDescription = function(html) {
...
+var aliasRegExp = /<a href="\/user\/([^"]+)/
+exports.getAuthor = function(body) {
+  var ownerinfo = this.between(body, '<div id="watch7-user-header" class=" spf-link ">', '<div id="watch8-action-buttons" class="watch-action-buttons clearfix">');
+  ownerinfo = new Entities().decode(ownerinfo)
+  if (ownerinfo === '') {
+    return {};
+  }
+  var match = ownerinfo.match(authorRegexp);
+  return {
+    ref: '/channel/'+match[1],
```
fent: Spaces around `+`

test/util-test.js
```diff
@@ -261,6 +276,65 @@ describe('util.getVideoDescription()', function() {
 });
 
 
+describe('util.getAuthor()', function() {
+  it('Retrieves formatted video author', function() {
+    var html = fs.readFileSync(path.resolve(__dirname,
+      'files/util/related-video'), 'utf8');
+    var authorObj = util.getAuthor(html);
+    assert.deepEqual(authorObj,{
+      type: 'channel',
```
fent: This was removed

TimeForANinja:
never expected this to end like that
thanks for spending this much time already
gonna add the example for the not existing user soon

lib/util.js
TimeForANinja: so u mean like `.match(/user/)` ?
wouldnt that ignore the slashes so that i would have to use `/\/user\//`

lib/info.js
```diff
@@ -144,7 +156,9 @@ function gotConfig(id, options, description, config, callback) {
     }
 
     info.formats = util.parseFormats(info);
-    info.description = description;
+
+    // Add additional properties to info.
+    info = util.objectAssign(info, additional)
```
fent: Add semicolon to end of line.

lib/util.js
```diff
@@ -291,6 +306,63 @@ exports.getVideoDescription = function(html) {
 
 
 /**
+ * Get video Owner from html.
+ *
+ * @param {String} body
+ * @return {Object}
+ */
+var authorRegexp = /<a href="\/channel\/([\w-]+)"[^>]+>(.+?(?=<\/a>))/
+var aliasRegExp = /<a href="\/user\/([^"]+)/
```
fent: End with semicolon

TimeForANinja:
`Also, how often does the user not get matched? Maybe a test could be added where it finds the channel but no user?` not to often ye. Thinking about wether i should add yet another file to the /test/files/util or edit the multiline description file

fent:
> just saw the util.assignDeep function and not sure wether util.objectAssign is actually neded just to get rid of one if()

Ah you're right! How did I forget about that D:

And yeah, looks like we can use that one. And use a boolean flag `deep` to decide if to assign recursively. For this PR for example, that isn't needed since everything in the `additional` object is created inside ytdl.

fent:
New version has been published. Thank you for all the hard work. :)

TimeForANinja:
Thank you very much for not just pressing no but helping me match it ur style
have to look into youtube's structure anyway for my api
also wanna do one more thing before im out for a while

Issue #102, opened 2017-01-27
Status code 404 - getting playlists


Heroj04:
Using getInfo() with a playlist URL returns the following error.
Is this not implemented?

using the youtube-dl package allows this feature to return an array with info of each video.

`
Error: Status code 404
    at ClientRequest.<anonymous> (C:\Users\jackf\Dropbox\Programming\Discord Bots\Neo\node_modules\ytdl-core\lib\request.js:33:18)
    at emitOne (events.js:96:13)
    at ClientRequest.emit (events.js:188:7)
    at HTTPParser.parserOnIncomingClient [as onIncoming] (_http_client.js:473:21)
    at HTTPParser.parserOnHeadersComplete (_http_common.js:99:23)
    at TLSSocket.socketOnData (_http_client.js:362:20)
    at emitOne (events.js:96:13)
    at TLSSocket.emit (events.js:188:7)
    at readableAddChunk (_stream_readable.js:176:18)
    at TLSSocket.Readable.push (_stream_readable.js:134:10)
`

fent:
It's not.

But it shouldn't return that error, it should instead say it was not able to find the video ID.

Heroj04:
Oh OK,

Are there any plans to implement this in the future?

And in regards to this specific error, It works frine with no error if I use a video URL.

fent:
No plans.

I'm not sure if this feature belongs in this module. If we include this, why not include getting videos from a user's uploads, a user's liked videos, etc.

There's also file to consider when adding an additional feature. Some use this module on the frontend, there's been [efforts](https://github.com/fent/node-ytdl-core/pull/38) to keep dependencies low.

karimsa:
I'm getting this error when trying to fetch a video, although it was working perfectly yesterday.

Code:

```javascript
var fs = require('fs')
    , ytdl = require('ytdl-core')

ytdl('http://youtube.com/watch?v=CovpoxCXCTw')
  .pipe(fs.createWriteStream('./audio.mp3'))
```

Stack trace:

```
Error: Status code 404
    at ClientRequest.<anonymous> (node_modules/ytdl-core/lib/request.js:33:18)
    at emitOne (events.js:96:13)
    at ClientRequest.emit (events.js:188:7)
    at HTTPParser.parserOnIncomingClient [as onIncoming] (_http_client.js:473:21)
    at HTTPParser.parserOnHeadersComplete (_http_common.js:99:23)
[...]

PR #107, opened 2017-02-01
Fixes Playlist throwing 404

TimeForANinja:
related to https://github.com/fent/node-ytdl-core/issues/102


lib/util.js
```diff
@@ -230,7 +230,7 @@ exports.getVideoID = function(link) {
     var s = parsed.pathname.split('/');
     id = s[s.length - 1];
   }
-  if (!id) {
+  if (!idRegex.test(id)) {
     throw new Error('No video id found: ' + link);
```
fent: Problem with this is if user gives a link when they accidentally didn't copy the last character of the id, they will get the error

    No video id found: http://www.youtube.com/watch?v=th75xfvBi5

Which looks like it's a correct link, but it's missing a character. Maybe a better solution is to still check for `if (!id)` and use this message then, and then also check if the id is set but did not match with the regexp, and have some other error for that.

Something like

`'Video id found (' + id + ') does not match expected format (' + idRegex + ')'`

test/util-test.js
```diff
@@ -222,16 +222,17 @@ describe('util.between()', function() {
...
+    assert(id, 'RAW_VIDEOID');
+    id = util.getVideoID('http://youtu.be/RAW_VIDEOID');
+    assert(id, 'RAW_VIDEOID');
+    id = util.getVideoID('http://youtube.com/v/RAW_VIDEOID');
+    assert(id, 'RAW_VIDEOID');
+    id = util.getVideoID('http://youtube.com/embed/RAW_VIDEOID');
+    assert(id, 'RAW_VIDEOID');
     id = util.getVideoID('RAW_VIDEOID'); // Video ids are 11-character long
     assert(id, 'RAW_VIDEOID');
+    assert.throws(function(){util.getVideoID('www.youtube.com/playlist?list=1337')}, Error, 'No video id found: www.youtube.com/playlist?list=1337');
```
fent: This line is a bit too long, make it a bit easier to read by adding some new lines inside the function.

In the last PR, although there were a few long lines, it seemed harder to split them up since they were html strings.

lib/util.js
```diff
@@ -233,6 +233,9 @@ exports.getVideoID = function(link) {
   if (!id) {
     throw new Error('No video id found: ' + link);
   }
+  if (!idRegex.test(id)) {
+    throw new Error('Video id ' + id + ' does not match expected format '+idRegex.toString());
```
fent: Spaces between `+`

What do you think of wrapping the `id` and `idRegex` in parenthesis in the string, to make it more clear where they start and end?

test/util-test.js
```diff
@@ -222,16 +222,21 @@ describe('util.between()', function() {
...
+    id = util.getVideoID('http://youtu.be/RAW_VIDEOID');
+    assert(id, 'RAW_VIDEOID');
+    id = util.getVideoID('http://youtube.com/v/RAW_VIDEOID');
+    assert(id, 'RAW_VIDEOID');
+    id = util.getVideoID('http://youtube.com/embed/RAW_VIDEOID');
+    assert(id, 'RAW_VIDEOID');
     id = util.getVideoID('RAW_VIDEOID'); // Video ids are 11-character long
     assert(id, 'RAW_VIDEOID');
+    assert.throws(
+      function () { util.getVideoID('www.youtube.com/playlist?list=1337') },
```
fent: New line before `,`

And the next three lines can be in one line.

```js
  assert.throws(function() {
    util.getVideo(...);
  }, Error, 'No video id found...');
```

test/util-test.js
```diff
@@ -222,16 +222,21 @@ describe('util.between()', function() {
...
+      Error,
+      'No video id found: www.youtube.com/playlist?list=1337'
+    );
   });
```
fent: An testing for an invalid id, like one with more or fewer characters, to test for that other error message.

lib/util.js
TimeForANinja: u mean like `'Video id(' + id + ') does not...'`?

test/util-test.js
```diff
@@ -222,16 +222,22 @@ describe('util.between()', function() {
...
+    assert(id, 'RAW_VIDEOID');
+    id = util.getVideoID('http://youtube.com/embed/RAW_VIDEOID');
+    assert(id, 'RAW_VIDEOID');
     id = util.getVideoID('RAW_VIDEOID'); // Video ids are 11-character long
     assert(id, 'RAW_VIDEOID');
+    assert.throws(function () {
+      util.getVideoID('www.youtube.com')
+    }, Error, 'No video id found: www.youtube.com');
+    assert.throws(function () {
+      util.getVideoID('www.youtube.com/playlist?list=1337')
```
fent: Add a semicolon at end of line on both of these.

fent:
Thanks again. New version published.

PR #111, opened 2017-02-05
Add the channel & video link to information

FireController1847:
It'd be awesome to be able to just get the channel link or video link by doing info.channel_link or info.video_link, so I propose this.

### DIFF
--- a/lib/info.js
+++ b/lib/info.js
@@ -202,8 +202,8 @@ function gotConfig(id, options, additional, config, callback) {
             });
           }
           info.formats.sort(util.sortFormats);
-          info.channel_link = (info.ucid ? `https://www.youtube.com/channel/${info.ucid}` : "No UCID");
-          info.video_link = (info.video_id ? `https://www.youtube.com/watch?v=${info.video_id}` : "No VIDEO_ID");
+          info.channel_link = (info.ucid ? "https://www.youtube.com/channel/" + info.ucid : "No UCID");
+          info.video_link = (info.video_id ? "https://www.youtube.com/watch?v=" + info.video_id : "No VIDEO_ID");
           callback(null, info);
         });
       });
--- a/lib/info.js
+++ b/lib/info.js
@@ -202,6 +202,8 @@ function gotConfig(id, options, additional, config, callback) {
             });
           }
           info.formats.sort(util.sortFormats);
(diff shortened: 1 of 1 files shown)


FireController1847:
I don't understand the issue. The code looks completely find and it worked when I was testing with it.

FireController1847:
Dammit, why wouldn't we use `` instead of "". If you still use 0.12 update your shit.

fent:
This feels unnecessary, since every video would have these prepended.

FireController1847:
But there doesn't seem to be an easy way to get the channel link or video link if you simply just want the link when getting information. It just seems kind of silly to require people to add the link in front of it

fent:
It doesn't seem that much more to prepend the YouTube domain to a channel id. As for the video URL, there are a few, but the main two are

    https://www.youtube.com/watch?v=VIDEO_ID
    https://youtu.be/VIDEO_ID

It's better to give the flexibility of choosing which one to use to link, rather than one or a list of them. There might be more in the future, like an even shorter one than `youtu.be`, maybe `y.tb`?

Also, if a user called this library, they likely already have the video link.

FireController1847:
The only reason I really wanted to add this is because I call the yt.getInfo functions many times and each time I have to add this because there's not an easy way to add the video & channel link to the information. I always have to add these two lines info.channel_link = bla bla each time I call it, and I call it: A lot.

fent:
Can I see an example or explanation where this is a common use case?

Also, it looks like we didn't need to parse the channel ID in #101

FireController1847:
Sure. Let me grab some images from my bot, one second.

FireController1847:
https://www.visual-fire.com/uploader/Images/Fire/53d58.png

I set the variables there, and then send an embed in the Discord chat with the video and channel link here.

https://www.visual-fire.com/uploader/Images/Fire/kkwiz.png

I also use it for the now playing command, as it gives people the video link and the url.
My bot is community built, so people have to submit songs to get them into the bot. I use this to give my submitters the video link so they can click on it and find the video right away. I always set these variables every time I use .getInfo, because I know I will always be using them.

https://www.visual-fire.com/uploader/Images/Fire/j5ct1.png

fent:
Published in v0.9.0

PR #112, opened 2017-02-05
Update README.md

FireController1847:


fent:
I meant another part of the readme where it says

    Info and format may look like this.

With a link to a gist of an info.

FireController1847:
@fent Sorry I'm so bad at this, github is not my strong point x3

Issue #81, opened 2016-07-10
Random ECONNRESET


MergHQ:
Hello,

I've had a issue with this library ever since I started to use it. I basically pass the ytdl stream into ffmpeg to extract the audio from the video. Sometimes the stream crashes with an ECONNRESET, and i have no idea why.

Stacktrace:
`Error: read ECONNRESET 
    at exports._errnoException (util.js:1007:11) at TLSWrap.onread (net.js:563:26)`

fent:
Do you have an example where this happens?

Is the example [here](https://github.com/fent/node-ytdl-core/blob/master/example/ffmpeg.js) similar?

MergHQ:
@fent [This](https://github.com/MergHQ/DiscordMusicBot/blob/master/src/commands/songrequests/Player.js) is where it happens.

MJunghenn:
Got that error, too. No specific Video or something just happened - and nothing can be downloaded.

fent:
There's no need to call `ytdlStream.end()` btw, `.destroy()` should be enough in theory.

hydrabolt:
Have been getting this error for a few months now...

MergHQ:
I have noticed that using the audioonly flag actually helps. Doesn't throw that error so often anymore.

fent:
Does this happen at the start of the download, middle, end?

hydrabolt:
Mainly the end

hydrabolt:
Towards the end

ShahramJ:
Happening to me too.

Brain-Gamer:
I have the same issue, but it happens to me in the middle / first half of the video

derekmartinez18:
I've noticed the issue occur sporadically, sometimes I can go a few videos with nothing while others cut off half way or near the end of playback.

Brain-Gamer:
Yeah I think it is pretty much random. But for me it's more often at the beginning of a video (most times)

FireController1847:
Can also report this issue. Happens randomly through playback, just started happening recently. I've always had this with connection loss, but the bot is currently not losing connection and it's just happening while the song was playing fine. Seems to be more common in the past few days.

FireController1847:
@fent People are saying it could be due to the fact you are downloading  videos "unofficially", or without a developer key.

TinyMan:
Same over here, random but mostly after the middle of the video

OzairP:
[...]

PR #128, opened 2017-03-19
continue after econnreset

TimeForANinja:
[x] restarting when the response stream ends before all data was received(up to 5 times)
[x] progress event since were already tracking it
[x] made content warnings easyer to adjust to each language
[x] kept the range options
[x] fixed tests
[x] fixed README

!! stream may emit 'response' multiple times
should also works as a workaround for #81

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -66,9 +66,9 @@ function downloadFromInfoCallback(stream, info, options) {
     trys: 5,
     range: {
       start: options.range ? (options.range.start ? options.range.start : 0) : 0,
-      end: options.range ? (options.range.end ? options.range.end : -1) : -1
+      end: options.range ? (options.range.end ? options.range.end : -1) : -1,
     },
-    downloaded: 0
+    downloaded: 0,
   });
 }
 
@@ -83,7 +83,7 @@ var redirectCodes = new Set([301, 302, 303, 307]);
  * @param {String} url
  * @param {Object} options
  * @param {Number} tryCount Prevent infinite redirects.
- * @param {Object} reconnectInfo
+ * @param {Object} reconnectInfo Continue after ECONNRESET
  */
 function doDownload(stream, url, options, tryCount, reconnectInfo) {
   if (stream._isDestroyed) { return; }
@@ -100,7 +100,7 @@ function doDownload(stream, url, options, tryCount, reconnectInfo) {
   var myrequest = options.request || request;
   var tempUrl = url;
   if (reconnectInfo.downloaded != 0 || reconnectInfo.range.start != 0 || reconnectInfo.range.end != -1) {
-    tempUrl += '&range=' + (reconnectInfo.range.start + reconnectInfo.downloaded) + '-' + (reconnectInfo.range.end != -1 ? reconnectInfo.range.end : '')
+    tempUrl += '&range=' + (reconnectInfo.range.start + reconnectInfo.downloaded) + '-' + (reconnectInfo.range.end != -1 ? reconnectInfo.range.end : '');
   }
   var req = myrequest(tempUrl, options.requestOptions);
 
--- a/lib/index.js
+++ b/lib/index.js
@@ -65,8 +65,8 @@ function downloadFromInfoCallback(stream, info, options) {
   doDownload(stream, url, options, 3, {
     trys: 5,
     range: {
-      start: options.range ? (options.range.start ? options.range.start : 0) : 0,
-      end: options.range ? (options.range.end ? options.range.end : -1) : -1,
+      start: options.range && options.range.start ? options.range.start : 0,
+      end: options.range && options.range.end ? options.range.end : -1,
     },
     downloaded: 0,
   });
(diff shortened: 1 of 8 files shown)


lib/index.js
```diff
@@ -112,15 +125,31 @@ function doDownload(stream, url, options, tryCount) {
...
+        doDownload(stream, res.headers.location, options, tryCount - 1, reconnectInfo);
         return;
       }
       stream.emit('response', res);
       stream.emit('error', new Error('Status code ' + res.statusCode));
       return;
     }
 
-    res.pipe(stream);
+    reconnectInfo.range.end = reconnectInfo.range.end === -1 ? parseInt(res.headers['content-length']) : (parseInt(res.headers['content-length']) < reconnectInfo.range.end ? parseInt(res.headers['content-length']) : reconnectInfo.range.end);
```
TimeForANinja: yep missed out on that...

lib/index.js
```diff
@@ -112,15 +125,31 @@ function doDownload(stream, url, options, tryCount) {
...
+    res.on('data', function(d) {
+      reconnectInfo.downloaded += d.length;
+      stream.emit('progress', reconnectInfo.downloaded / (reconnectInfo.range.end - reconnectInfo.range.start), reconnectInfo.range.start, reconnectInfo.range.end, reconnectInfo.downloaded, d.length);
```
TimeForANinja: not hurting since u can just ignore them in the listener if u only care about the percentage

lib/info.js
```diff
@@ -18,6 +18,10 @@ var KEYS_TO_SPLIT = [
   'fexp',
   'watermark'
 ];
+var CONTENT_WARNINGS = new Set([
+  'Content Warning',
+  'Inhaltswarnung'
```
TimeForANinja: Guess i missed out deleting that again. Was a little fix 'cause the test's kept failing

lib/index.js
```diff
@@ -112,15 +125,31 @@ function doDownload(stream, url, options, tryCount) {
...
+    });
+
+    res.pipe(stream, {end : false});  //not sure wether this pipes errors...
+    res.on('end', function() {
+      stream.unpipe(res);
+      if (reconnectInfo.downloaded + reconnectInfo.range.start < reconnectInfo.range.end) {
```
TimeForANinja: u mean just like 2 streams piped into another with one emitting an error and the otherone listening for it?

lib/index.js
fent: Well, we also don't want to overpopulate the API. I'm not sure that including the `range` would reveal new information to the user. If the `range` option isn't used, then the range here is set to the default and that's what they would get. If they do use the `range` option, they would get back what they used.

lib/index.js
fent: I had thought that the reason `end` was being listened to was because it was expected to happen when there stream suddenly disconnects mid-download. But if it does disconnect, an `error` event would be emitted, but not an `end` event.

lib/index.js
```diff
@@ -133,15 +133,22 @@ function doDownload(stream, url, options, tryCount, reconnectInfo) {
       return;
     }
 
-    reconnectInfo.range.end = reconnectInfo.range.end === -1 ? parseInt(res.headers['content-length']) : (parseInt(res.headers['content-length']) < reconnectInfo.range.end ? parseInt(res.headers['content-length']) : reconnectInfo.range.end);
+    if(reconnectInfo.downloaded === 0) {
```
fent: Space after `if`

lib/index.js
```diff
@@ -133,15 +133,22 @@ function doDownload(stream, url, options, tryCount, reconnectInfo) {
...
+      var res_length = parseInt(res.headers['content-length']);
+      var end = reconnectInfo.range.end;
+      var shortest_length = Math.min(res_length, end);
+      reconnectInfo.range.end = end === -1 ? res_length : shortest_length;
```
fent: Please use camelCase to keep project code a bit more consistent.

lib/info.js
TimeForANinja: Gonna delete it in here and open an issue until someone finds a better way...

lib/index.js
TimeForANinja: It should be emitted, yes. Not happening everytime thou as far as i know.

fent: How are you testing it?

TimeForANinja: Had a video that ended mid way throu every time. Just added logs for all events i could think / read off in the index.js and ran it a few times and played with it. Unluckily i had to reset my pc few days ago and lost that file including my script and the videos

TimeForANinja: with my fix https://hastebin.com/yukifuhixi.coffeescript
the index.js dodownload func used in that https://hastebin.com/xezuwinita.js

without my fix https://hastebin.com/haquwunafo.coffeescript
again the dodownload func https://hastebin.com/fimakeziha.js

fent: In the first test, how was the stream aborted halfway?

TimeForANinja: Looks.like the res just finished

fent:
I haven't forgotten about this, I checked out the branch locally and added some tests. But I won't be finished for a couple of days.

fent:
Closing in favor of #139

PR #137, opened 2017-03-30
Add Typescript definition file

iCrawl:
Feels like the old typescript PR has been stagnating for a bit.
Typescript 2.x has been released in the meantime too, so there is no reason to use the old `typings` anymore.
NPM support for typings improved too.

All of this means there is less changes needed in the source files you already provide except:

- Add types entry to package.json (It doesn't matter whether its `typings` or `types` both are fine to use)
- Add node typings to the devDependencies (This is important to import ClientRequest from http and Readable from stream, since those are native node modules)
- - This won't affect any non-typescript user, but makes life easier for those who choose to write in TS.
- Add typescript devDependency to have an easier time updating and testing the types
- Add very simple tsconfig.json (https://www.typescriptlang.org/docs/handbook/tsconfig-json.html)


package.json
```diff
@@ -24,13 +25,15 @@
...
+    "@types/node": "^7.0.12",
     "assert-diff": "^1.0.1",
     "istanbul": "*",
     "mocha": "*",
     "muk-prop": "^1.0.0",
     "nock": "*",
     "sinon": "^2.0.0",
-    "stream-equal": "~1.0.0"
+    "stream-equal": "~1.0.0",
+    "typescript": "^2.2.2"
```
fent: Does this need to be included? In another recent similar PR (for another project), `typescript` was not part of `devDependencies`.

Which does make sense since someone working with TS would already have that on their project.

typings/index.d.ts
```diff
@@ -0,0 +1,76 @@
...
+    sendtmp: number;
+    title: string;
+    status: string;
+    muted: number;
+    pltype: string;
+    thumbnail_url: string;
+    has_cc: boolean;
+    video_verticals: number[];
+    track_embed: number;
+    author: string;
```
fent: There's been some changes since the last PR, `author` is now an object that looks like [this](https://github.com/fent/node-ytdl-core/blob/master/example/info.json#L51)

And it also now includes [`published` and `relatedVideos`](https://github.com/fent/node-ytdl-core/blob/master/example/info.json#L620)

package.json
iCrawl: This could be debatable I agree, but since those are devDependencies, and they are for working on `this` specific module, you could also argue mocha and instanbul are not needed, since those are test engines used in many popular projects already.
But you obviously need those 2 to run tests whether through travis or offline when you add new features.
If you want me to remove it, its fine though.

typings/index.d.ts
iCrawl: I will add those, thanks!

iCrawl: Looks like the info.json seems to be a bit more outdated, I just requested the video you put into `progress.js` and I get a lot of more info then what info.json reflects.
I will update it with those values.

package.json
fent: Mocha and istanbul are used directly by this module. But it doesn't use TypeScript directly, it doesn't need it to run tests on travis for example, or when developing for the module.

iCrawl: Anyone wanting to add typescript stuff to this module (updating def. file for example) would need to have a global installation of it, or install it locally without saving it to the package.json.

But like I said, I am fine removing it, your call.

fent: It's not a common situation and it's an additional big dependency during tests. I think it would be fine to remove it.

iCrawl:
I hope I didn't go overboard with the types here, haha.

I checked 5 different videos, from new to old, 8K to non-HD and marked the entries depending on that.
If someone is only on newer or older videos its prefixed with a `?` which means it may or may not be there when you retrieve the information.

typings/index.d.ts
```diff
@@ -0,0 +1,193 @@
...
+  import { ClientRequest } from 'http';
+  import { Readable } from 'stream';
+
+  export function ytdl(link: string, options?: downloadOptions): Readable;
+  export namespace ytdl {
+    export function getInfo(url: string, callback?: (err: Error, info: videoInfo) => void): Promise<videoInfo>;
+    export function getInfo(url: string, options?: downloadOptions, callback?: (err: Error, info: videoInfo) => void): Promise<videoInfo>;
+    export function downloadFromInfo(info: videoInfo, options?: downloadOptions): Readable;
+    export function chooseFormat(format: object[], options?: downloadOptions): object | Error;
+    export function filterFormats(formats: object[], filter?: 'video' | 'videoonly' | 'audio' | 'audioonly' | ((format: videoFormat) => boolean)): object[];
```
fent: Shouldn't these two `format: object[]` be `format: videoFormat[]`?

iCrawl: Right, damn. Sorry.

fent:
Thanks for going through a number of videos and listing out all those properties for this PR 👍

PR #139, opened 2017-04-02
Continue download after early end

fent:
Continuing from #128, this is mainly adding tests.

I also tested this with a real download, by pausing a download halfway for about 6 minutes, I was able to consistently get an ECONNRESET error and then continue the download.

@TimeForANinja please look over this, and feel free comment on anything I might have overlook.

### DIFF
--- a/example/range.js
+++ b/example/range.js
@@ -5,7 +5,7 @@ var ytdl = require('..');
 var url = 'https://www.youtube.com/watch?v=CehAKQL463M';
 var output = path.resolve(__dirname, 'video.mp4');
 
-var video = ytdl(url, { range: '0-1000' });
+var video = ytdl(url, { range: { start: 0, end: 1000 } });
 video.pipe(fs.createWriteStream(output));
 
 video.on('end', function() {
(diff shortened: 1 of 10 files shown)


example/range.js
```diff
@@ -5,10 +5,10 @@ var ytdl = require('..');
 var url = 'https://www.youtube.com/watch?v=CehAKQL463M';
 var output = path.resolve(__dirname, 'video.mp4');
 
-var video = ytdl(url, { range: '0-1000' });
+var video = ytdl(url, { range: { start: 0, end: 1000 } });
 video.pipe(fs.createWriteStream(output));
 
 video.on('end', function() {
-  ytdl(url, { range: '1001-' })
+  ytdl(url, { range: { start: 0 } })
```
TimeForANinja: Isn't this suppost to be 1000?

lib/index.js
```diff
@@ -79,30 +83,48 @@ var redirectCodes = new Set([301, 302, 303, 307]);
...
       myres.unpipe();
+      myres.destroy();
     }
     stream.emit('abort');
   };
 
   req.on('error', function(err) {
-    stream.emit('error', err);
+    if (stream._isDestroyed || !myres) {
+      stream.emit('error', err);
```
TimeForANinja: Not sure wether u shouldnt close the stream anyway but since i don't know better...

fent: The stream should close itself after this.

fent:
Thanks for the original PR again, and for reviewing.

PR #142, opened 2017-04-04
Update Info.json

FireController1847:
Update Info.json to the latest version of Ytdl.


fent:
Thanks for the PR.

Your ip is left in there btw. I usually replace it with `ip=0.0.0.0` but if you don't care I'll merge as is.

Hmm, YouTube seems to change the properties it returns every few days. There's only so much we can do to guarantee certain attributes.

FireController1847:
@fent Yeah. I always forget my IP is in those things haha, thanks for reminding me to remove it.

fent:
You might wanna reset before the last two commits (without undoing the changes), then commit again, and then force push. Otherwise the commits that add/remove your ip will still be there.

FireController1847:
@fent It'll be fine, when I tried to match the IP with the one google gave it was different so, whatever.

fent:
Alright cool. Thanks again.

FireController1847:
No problem! :D

PR #143, opened 2017-04-09
make error messages international

### LINKED ISSUE #134: Missing Language support for age restricted videos
This ignores other languages besides english
https://github.com/fent/node-ytdl-core/blob/master/lib/info.js#L60

TimeForANinja:
by changing the language for the info request to english
+ fix tests
+ fix #134

### DIFF
--- a/lib/info.js
+++ b/lib/info.js
@@ -48,7 +48,7 @@ module.exports = function getInfo(link, options, callback) {
   if (id instanceof Error) return callback(id);
 
   // Try getting config from the video page first.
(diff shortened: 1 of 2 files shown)


fent:
I did not know about this parameter. But when I try going to a url with it, I get the last language I set. But I also get a message asking if I want to switch to the language from the param.

fent:
I found something else that works, will push shortly. Thanks anyway for the efforts 👍

TimeForANinja:
All fine 👍

PR #148, opened 2017-05-01
Fix typings

DarkerTV:
Currently the type definitions file is not correctly exporting the `ytdl` function and its own namespace making it impossible to use in typescript.

This PR fix this export isolating everything inside the `ytdl` namespace and exporting only the main function with the namespace, already tested and working in a local project.


fent:
I'm not too familiar with using TypeScript, how can I properly test this? If I create a file requiring ytdl-core, and call it in a wrong way, I don't get a compile error from `tsc`.

@iCrawl, you added the last typings file, does this change look good?

iCrawl:
Hello,

Yes, I don't see the issue with how it is currently running.



I guess one could add `export default ytdl;` right below `export function ytdl [...]` to allow `import ytdl from 'ytdl-core';` instead of destructuring, but thats just icing on the cake right there.

DarkerTV:
@iCrawl You don't get any error while writing because the typings file is "valid" for typescript. I get the error at compile time `ytdl_core_1.ytdl is not a function`.
image

This is because https://github.com/fent/node-ytdl-core/blob/master/lib/index.js#L14 use `module.exports = function ytdl`, so ytdl is the default export, not a membed of exports.

Also you can't use `import ytdl from 'ytdl-core';` because this is a sugar syntactic for `const ytdl = require('ytdl-core').default` and you can't currently import ytdl-core this way.

iCrawl:
LGTM @fent 👍

fent:
Thanks for the work!

PR #164, opened 2017-05-25
-Fixes downloading of specific video issue

### LINKED ISSUE #163: error 500 when trying to download a specific video
Hello there,

I am trying to move away from youtube-dl for a specific issue I found and also due to the author not actively maintaining the lib anymore.

I see your project is getting updated so i wanted to test it out.

Unfortunately I found out the same issue on your project which I invite you to try.

Try downloading http://www.youtube.com/watch?v=fxlbIS8CoW8 using best quality (by default)

```
const fs = require('fs');

ytdl('http://www.youtube.com/watch?v=fxlbIS8CoW8')
  .pipe(fs.createWriteStream('video.flv'));
```

This should suffice to break the library. for some unknown reasons this specific video and several others will not work.

I am downloading thousand of vidoes for a project and while this might happen rarely, it happens quite a lot for me.

Let me know if you need further details to reproduce the bug.

Hopefully you'll find the reason, i wasn't able to.

Best regards,
Nikooo777

cacheflow:
-Quality defaults to highest in chooseFormat function when users do not pass a quality option.
-Sometimes the highest quality video has an invalid or inaccessible URL that cannot be downloaded.
-This fixes the issue by performing a request to the URL and setting a fallback quality option on failure
so a user can still download a video.
-Also, included a new format called baseline.
-Resolves #163.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -60,30 +60,38 @@ function downloadFromInfoCallback(stream, info, options) {
   }
   stream.emit('info', info, format);
   if (stream._isDestroyed) { return; }
-
   var url = format.url;
-  if (format.live) {
-    var req = m3u8stream(url, {
-      chunkReadahead: +info.live_chunk_readahead,
-      requestOptions: options.requestOptions,
-    });
-    req.on('error', stream.emit.bind(stream, 'error'));
-    stream.destroy = req.end.bind(req);
-    req.pipe(stream);
 
-  } else {
-    if (options.begin) {
-      url += '&begin=' + util.fromHumanTime(options.begin);
+  request(url, {}, function(err, res, body) {
+    if (err) {
+      format = util.chooseFormat(info.formats, {quality: 'baseline'});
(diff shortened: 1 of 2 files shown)


cacheflow:
Brb fixing failing tests.

fent:
This doesn't actually fix the issue, it instead downloads a different video if there's an error with the first quality tried.

Although having the option to fallback to another quality would be a better way to do it.

cacheflow:
I assumed the issue wasn't within node-ytdl, but on YouTube's because I had the same result with another module.

fent:
If you can watch the video in your browser with the same quality, then it's an issue with this module.

cacheflow:
Good point.

fent:
Closing this for now.

I'm not sure what the `baseline` profile for a format is. If you believe we should have this filter, feel free to open another PR for it.

cacheflow:
Definitely.

PR #175, opened 2017-06-10
Request for iurlmaxres (Fixes #155) (Fix #1)

### LINKED ISSUE #155: How does one check if the video supports maxresdefault?

FireController1847:
This request should be able to fix the issue #155, by sending a request. YouTube, by default, sends a 404 status code response if it doesn't exist, and a 200 response if it does. By doing this, we will set iurlmaxres if it gets a 404 code.

If you want, I can remove the .message.includes and just have it set it null if it errored at all, but I think this may be the best method for what we want.

### DIFF
--- a/lib/info.js
+++ b/lib/info.js
@@ -65,7 +65,7 @@ module.exports = function getInfo(link, options, callback) {
     
     // Check if video supports maxresdefault.
     var iurlmaxres = THUMBNAIL_URL + id + '/maxresdefault.jpg';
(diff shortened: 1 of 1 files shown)


FireController1847:
I don't understand what the error is.

FireController1847:
I've got to go to bed will look at it more tomorrow <3

FireController1847:
Let's just drop this subject, and I'll add something to the readme instead to give users a heads up.

PR #176, opened 2017-06-11
Request for iurlmaxres (Fixes #155) (Fix #2)

### LINKED ISSUE #155: How does one check if the video supports maxresdefault?

FireController1847:
This request should be able to fix the issue #155, by sending a request. YouTube, by default, sends a 404 status code response if it doesn't exist, and a 200 response if it does. By doing this, we make iurlmaxres a function, that way it only makes the request when they want it, and it will return null if 404.

If you want, I can remove the .message.includes and just have it set it null if it errored at all, but I think this may be the best method for what we want.

### DIFF
--- a/lib/info.js
+++ b/lib/info.js
@@ -84,7 +84,17 @@ module.exports = function getInfo(link, options, callback) {
       iurlsd     : THUMBNAIL_URL + id + '/sddefault.jpg',
       iurlmq     : THUMBNAIL_URL + id + '/mqdefault.jpg',
       iurlhq     : THUMBNAIL_URL + id + '/hqdefault.jpg',
-      iurlmaxres : THUMBNAIL_URL + id + '/maxresdefault.jpg',
+      iurlmaxres : function getiurlmaxres(iurlcallback) {
+        var iurlmaxres = THUMBNAIL_URL + id + '/maxresdefault.jpg';
+        request(iurlmaxres, function(iurlerr, iurlres, iurlbody) {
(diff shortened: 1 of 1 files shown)


lib/info.js
```diff
@@ -84,7 +84,21 @@ module.exports = function getInfo(link, options, callback) {
       iurlsd     : THUMBNAIL_URL + id + '/sddefault.jpg',
       iurlmq     : THUMBNAIL_URL + id + '/mqdefault.jpg',
       iurlhq     : THUMBNAIL_URL + id + '/hqdefault.jpg',
-      iurlmaxres : THUMBNAIL_URL + id + '/maxresdefault.jpg',
+      iurlmaxres : function getiurlmaxres(iurlcallback) {
+        var iurlmaxres = THUMBNAIL_URL + id + '/maxresdefault.jpg';
+        request(iurlmaxres, function(iurlerr, iurlres, iurlbody) {
+          if (iurlerr && iurlerr.message.includes("Status code: 404")) {
+            if (!iurlcallback) {
+              return new Promise(function(r, j) { return r(null)});
```
TimeForANinja: This is not how async works.
U wanna return a Promise wrapping the function before u do an async task like the request()

TimeForANinja: Also u should not resolve with null, rather reject

FireController1847: We WOULD want to resolve with null, because all we're doing is getting it.

FireController1847: In fact, it might be better just to remove promises at all.

PR #177, opened 2017-06-11
IURLMAXRES Readme headsup (Disregards #155)

FireController1847:


fent:
Hi, thanks for you contribution efforts.

Maybe I'm underestimating the importance of this property, but I don't think it's used widely to justify the sole note about it after the `info` description.

FireController1847:
I feel like it should, just to note that iurlmaxres will not ALWAYS be available. From my experience (on Discord), most users who use ytdl-core use iurlmaxres, and I've had a couple ask me about this situation. I can fully understand if you don't want to add it but I feel like enough people use it that it'd be nice to leave a note or something to say that it won't always be available, instead of just... leaving it, then users will complain over the fact "hey why isn't it working".

fent:
> From my experience (on Discord), most users who use ytdl-core use iurlmaxres

Do they use it for a discord bot? If so, `iurlhq` should suffice for a thumbnail.

It just feels odd to have that one note about a seemingly arbitrary property. Ideally, we'd document the entirety of the `info` object, but there's tons of stuff there, and it changes all the time. We could still try with the more consistent properties and the ones that are added by ytdl-core.

If that is done, I think the [typings file](https://github.com/fent/node-ytdl-core/blob/master/typings/index.d.ts) would be a good place to add comments. It's kept decently up to date with all properties from `info`.

FireController1847:
Personally, the black bars on my bot is actually quite annoying.



And, tbh, I don't even know about the typings. I don't know how many people read those but I guess we could somehow redirect users to it. (I'm not very familiar with TypeScript so you or someone who knows how would have to do it) I know I've never read the typings.

fent:
Ah, I didn't realize `iurlhq` had those black bars. If it helps, looks like `iurlmq` does not have them.

I don't use TypeScript either, but a typings file do a good job of showing the structure of objects. And yeah, we'd link users to it just like the `example/info.json` file is linked.

FireController1847:
@fent Idk it's your node module, up to you. Just tell me and I'll close em' both.

fent:
The other one's fine for now. I'll close this one.

Issue #165, opened 2017-05-27
Drop support for outdated node versions


TinyMan:
According to https://github.com/nodejs/LTS#lts-schedule1, v0.10 v0.12 are in End-of-Life and 4.x is in maintenance. 

With node 8.x around the corner it may be the right time to drop support for older unstable versions.

fent:
A while ago I thought which versions I should keep supporting. I came across [this source](https://semaphoreci.com/blog/2016/10/12/nodejs-versions-used-in-commercial-projects-2016-edition.html) that lists v0.12 as still being used substantially in 2016. Also, I didn't find it too limiting to not use some of the new things from v4. This is a decision made across all of the open source projects I maintain.

But, I can see how it could be annoying to want to contribute to a project, and not being able to use some es6 stuff.

TinyMan:
Ha I was not expecting this answer.

I was amazed to see how v0.12 is still used in 2016. I cannot understand why people do that ... It's like those folks still using windows xp and doing no updates that Microsoft has to save because of some monthly cyber attack even if xp is no longer maintained. I understand it's a question of money for them (upgrading can be difficult and time taker) but still, windows xp ...

It's not only about those stuff of es6, but also about the good habits. Also, I think that if we keep older versions alive, they will never be dropped. Few years later we end up using thousands of different versions all across the web and it breaks things. And from what I know there is no problem running v0.12 code into newer releases. Over here I user v6.x and never had a problem running `ytdl` for example. What I'm saying is that people can still use their older dependencies with newer Node version.

If someone needs `ytdl` there are 2 cases:
* either they use stable Node version and it runs, and it would still run with `ytdl` supporting only stable / maintenance versions
* or they use old Node version. They have the choice to update their Node version or to shrinkwrap their dependency and not get new features/fix from `ytdl` as they do for Node. I cannot see where someone would actively develop anything today with only support for End-of-Life Node versions. Maintenance is another thing and they should not be expecting new features from dependencies.

That being said, I know that doing updates can sometimes break things up.

Anyway, I can understand your choice even if I wouldn't have done the same. It's up to you. You can close this if you chose to wait.

fent:
> It's not only about those stuff of es6, but also about the good habits. Also, I think that if we keep older versions alive, they will never be dropped. Few years later we end up using thousands of different versions all across the web and it breaks things. And from what I know there is no problem running v0.12 code into newer releases.

[...]

PR #184, opened 2017-07-03
improve consistency in examples

TimeForANinja:
Tryed to stick with never syntax even thou i know about #165 since most examples already used things like `const` or arrow functions.


example/get_info.js
```diff
@@ -3,16 +3,16 @@ const fs   = require('fs');
 const ytdl = require('..');
 const id  = '7wNb0pHyGuI';
 
+const filepath = path.resolve(__dirname, 'info.json');
 
-ytdl.getInfo(id, function(err, info) {
+ytdl.getInfo(id, (err, info) => {
```
TimeForANinja: not sure wether i should use promises here for sticking to new syntax or the callback since the promise is just a wrapper

fent: This way is fine.

fent:
Using `.on` instead of `.once` is fine for events that should only be emitted once. Even in node API pages, `.on` is preferred. If there ever is another event, that is an error. `.once` is meant to be used more for events that happen more than once, but you're only interested in the first.

example/progress.js
```diff
@@ -2,16 +2,25 @@ const path = require('path');
...
+  starttime = Date.now();
+})
+video.on('progress', (chunkLength, downloaded, total) => {
+  const float_downloaded = downloaded / total;
+  const downloaded_min = (Date.now() - starttime) / 1000 / 60;
   process.stdout.cursorTo(0);
   process.stdout.clearLine(1);
-  process.stdout.write((downloaded / total * 100).toFixed(2) + '% ');
+  process.stdout.write(`${(float_downloaded * 100).toFixed(2)}% downloaded`);
+  process.stdout.write(`(${downloaded / 1024 / 1025}MB of ${total / 1024 / 1024}MB)`);
```
fent: Downloaded is not defined.

example/progress.js
```diff
@@ -2,16 +2,25 @@ const path = require('path');
...
+  process.stdout.write(`, running for: ${download_time.toFixed(2)}minutes`);
+  process.stdout.write(`, estimated time left: ${(download_time / float_downloaded - download_time).toFixed(2)}minutes`);
```
fent: The one line of this makes it so it is printed in one line, then erased, and printed again. It would show a progressive percentage updated in one line.

Adding extra lines changes this.

TimeForANinja: thought about extra lines but since i didn't know how to clear them again i just stuffed all in one
works fine thou and i guess making it pretty would be 2 much effort

example/progress.js
TimeForANinja: it is, gets passed into the function

TimeForANinja:
I guess they prefer `.on` since it uses less memory ¯\_(ツ)_/¯
i'll change it real quick even thou i'd consider this hole thing bad practise

example/progress.js
fent: Sorry, meant `download_time`

TimeForANinja: fixed

fent: This line needs a `.toFixed(2)` too

example/progress.js
```diff
@@ -2,16 +2,25 @@ const path = require('path');
...
+const output = path.resolve(__dirname, 'video.mp4');
 
-var video = ytdl(url);
+const video = ytdl(url);
+let starttime;
 video.pipe(fs.createWriteStream(output));
-video.on('progress', function(chunkLength, downloaded, total) {
+video.once('response', () => {
+  starttime = Date.now();
+})
```
fent: Semicolon at end of line

example/progress.js
```diff
@@ -2,16 +2,25 @@ const path = require('path');
...
+video.on('progress', (chunkLength, downloaded, total) => {
+  const float_downloaded = downloaded / total;
+  const downloaded_min = (Date.now() - starttime) / 1000 / 60;
```
fent: Would prefer these variables to be in camelCase, as the rest of the project is.

example/progress.js
```diff
@@ -2,16 +2,25 @@ const path = require('path');
...
   process.stdout.cursorTo(0);
   process.stdout.clearLine(1);
-  process.stdout.write((downloaded / total * 100).toFixed(2) + '% ');
+  process.stdout.write(`${(float_downloaded * 100).toFixed(2)}% downloaded`);
+  process.stdout.write(`(${downloaded / 1024 / 1025}MB of ${total / 1024 / 1024}MB)`);
+  process.stdout.write(`, running for: ${downloaded_min.toFixed(2)}minutes`);
+  process.stdout.write(`, estimated time left: ${(downloaded_min / float_downloaded - downloaded_min).toFixed(2)}minutes`);
```
fent: Since this line is really long, it won't be printed in one line some terminals if there isn't enough room. It will instead print everything in several new lines.

I did some research on how to continuously print to multiple lines, here's something that worked with the [readline module](https://nodejs.org/api/readline.html):

```js
  readline.cursorTo(process.stdout, 0);
  process.stdout.write(`${(float_downloaded * 100).toFixed(2)}% downloaded `);
  process.stdout.write(`(${downloaded / 1024 / 1025}MB of ${total / 1024 / 1024}MB)\n`);
  process.stdout.write(`running for: ${downloaded_min.toFixed(2)}minutes`);
  process.stdout.write(`, estimated time left: ${(downloaded_min / float_downloaded - downloaded_min).toFixed(2)}minutes`);
  readline.moveCursor(process.stdout, 0, -1);
```

TimeForANinja: works fine besides i need to add a space at the end of the last line else it prints `minutess` after some time

example/progress.js
```diff
@@ -1,17 +1,27 @@
...
+  starttime = Date.now();
 });
-video.on('end', function() {
-  process.stdout.write('\n');
+video.on('progress', (chunkLength, downloaded, total) => {
+  const floatDownloaded = downloaded / total;
+  const downloadedMinutes = (Date.now() - starttime) / 1000 / 60;
+  readline.cursorTo(process.stdout, 0);
+  process.stdout.write(`${(floatDownloaded * 100).toFixed(2)}% downloaded`);
+  process.stdout.write(`(${(downloaded / 1024 / 1025).toFixed(2)}MB of ${(total / 1024 / 1024).toFixed(2)}MB)\n`);
```
fent: Was wondering why the total downloaded never equaled the total when at 100%. There's a `1025` here instead of `1024`

example/progress.js
```diff
@@ -1,17 +1,27 @@
...
+  process.stdout.write(`running for: ${downloadedMinutes.toFixed(2)}minutes`);
+  process.stdout.write(`, estimated time left: ${(downloadedMinutes / floatDownloaded - downloadedMinutes).toFixed(2)}minutes `);
+  readline.moveCursor(process.stdout, 0, -1);
 });
+video.on('end', () => {
+  process.stdout.write('\n');
```
fent: Another `\n` needed since we're printing two lines now

fent:
Thanks again!

TimeForANinja:
Always welcome

PR #186, opened 2017-07-11
add validateLink

TimeForANinja:


lib/index.js
```diff
@@ -32,6 +32,17 @@ ytdl.filterFormats = util.filterFormats;
 ytdl.cache = cache;
 
 
+/**
+ * checks wether the input string includes a valid id
+ *
+ * @param {String} string
+ */
+ytdl.validateLink = function(string) {
```
fent: Would rather place this inside of `util.js` and export it here like the other util functions.

lib/index.js
```diff
@@ -32,6 +32,17 @@ ytdl.filterFormats = util.filterFormats;
...
+  if (util.getVideoID(string) instanceof Error) return false;
```
fent: Can return the result of `instanceof` negated.

```js
return !(util.getVideoID(string) instanceof Error);
```

TimeForANinja:
also thought about test but then since there are already tests for util.getVideoID i didn't

lib/index.js
```diff
@@ -30,6 +30,7 @@ ytdl.getInfo = getInfo;
 ytdl.chooseFormat = util.chooseFormat;
 ytdl.filterFormats = util.filterFormats;
 ytdl.cache = cache;
+ytdl.validateId = util.validateLink;
```
fent: Would prefer if this is one line above so it can be grouped with the other functions from `util`

test/util-test.js
```diff
@@ -305,6 +305,28 @@ describe('util.getVideoID()', function() {
...
+    assert.equal(id, true);
+    id = util.validateLink('RAW_VIDEOID'); // Video ids are 11-character long
+    assert.equal(id, true);
+    id = util.validateLink('https://www.twitch.tv/user/v/1234');
+    assert.equal(id.message, false);
+    id = util.validateLink('www.youtube.com');
+    assert.equal(id.message, false);
+    id = util.validateLink('http://www.youtube.com/playlist?list=1337');
+    assert.equal(id.message, false);
+  });
```
fent: No need to test for all cases, since they're all tested on the above test anyway. Just one that returns `true` and one that returns `false`.

test/util-test.js
```diff
@@ -305,6 +305,16 @@ describe('util.getVideoID()', function() {
   });
 });
 
+describe('util.validateLink()', function() {
+  it('Retrieves wether a string includes a parsable video ID', function() {
```
FireController1847: Whether. Whether, man.

TimeForANinja: noone saw that...

fent:
One more thing, can you add `validateLink()` to the readme?

After `ytdl.filterFormats()` should be good. Something like "Returns true if YouTube link is valid."

fent:
Thanks again!

PR #195, opened 2017-08-03
make regexp match more newline types

Marjoe:
make regexp match more newline types

### DIFF
--- a/lib/sig.js
+++ b/lib/sig.js
@@ -113,7 +113,7 @@ var actionsObjRegexp = new RegExp(
     jsKeyStr + sliceStr   + '|' +
     jsKeyStr + spliceStr  + '|' +
     jsKeyStr + swapStr +
-  '),?(\\r\\n|\\r|\\n)?)+)\\};'
+  '),?\\r?\\n?)+)\\};'
 );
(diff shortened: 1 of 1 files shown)


lib/sig.js
```diff
@@ -113,7 +113,7 @@ var actionsObjRegexp = new RegExp(
     jsKeyStr + sliceStr   + '|' +
     jsKeyStr + spliceStr  + '|' +
     jsKeyStr + swapStr +
-  '),?\\n?)+)\\};'
+  '),?(\\r\\n|\\r|\\n)?)+)\\};'
```
fent: Could be simplified as `\\r?\\n?`

Marjoe: I ran the tests on Windows 10 and failed. Then when I debug the code I found the 'body' in function extractActions in sig.js has all newline as \\r\\n. So I think this is caused by Windows.

fent: Yes but, wouldn't `\\r?\\n?` also match that string? Could you test that on Windows?

Marjoe: `\\r?\\n?` works well. Thanks for your guide:)

fent:
Thank you!

PR #201, opened 2017-08-11
[Optimize] ID should be 11 number

zhuyifan2013:
Some url could be played such as : 
`https://www.youtube.com/watch?v=inRtVJMVlBk=lru?`
`https://www.youtube.com/watch?v=inRtVJMVlBk=XXXX`
 , but this lib could not parse successful, so I suggest we could cut the id to 11 length.

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -236,7 +236,7 @@ exports.getVideoID = function(link) {
   if (!id) {
     return new Error('No video id found: ' + link);
   }
-  id = id.substring(0,11);
+  id = id.substring(0, 11);
   if (!idRegex.test(id)) {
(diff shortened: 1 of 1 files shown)


lib/util.js
```diff
@@ -236,6 +236,7 @@ exports.getVideoID = function(link) {
   if (!id) {
     return new Error('No video id found: ' + link);
   }
+  id = id.substring(0,11);
```
fent: Thanks for the PR!

Could you add a space after the comma? Trying to keep the styling consistent across files.

zhuyifan2013: Done.

fent:
Thanks again!

PR #202, opened 2017-08-12
Change "once" to "on"

Jqmey:


fent:
Why? Does this event happen more than once if say the client reconnects?

FireController1847:
@fent Yeah, it does.

PR #203, opened 2017-08-14
Fix error message for videos that require purchase

FD-:
requires_purchase seems to only be reported if the request sets the el parameter to info. It seems the checks for requires_purchase were never reached before, so the error message was unspecific.

### DIFF
--- a/lib/info.js
+++ b/lib/info.js
@@ -136,6 +136,7 @@ function gotConfig(id, options, additional, config, appendConfig, callback) {
       gl: 'US',
       hl: (options.lang || 'en'),
       sts: config.sts,
+      el: 'info',
     },
   });
   request(url, options.requestOptions, function(err, res, body) {
--- a/lib/info.js
+++ b/lib/info.js
@@ -142,11 +142,11 @@ function gotConfig(id, options, additional, config, appendConfig, callback) {
   request(url, options.requestOptions, function(err, res, body) {
     if (err) return callback(err);
     var info = querystring.parse(body);
-    if (info.status === 'fail') {
-      info = config.args;
-    } else if (info.requires_purchase === '1') {
+    if (info.requires_purchase === '1') {
       return callback(new Error(info.ypc_video_rental_bar_text));
-    }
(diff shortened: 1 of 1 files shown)


fent:
I've tested this with video id `SyKPsFRP_Oc` but it still gives `This video is unavailable`

Is there another video where this gives an expected error message?

FD-:
I was able to confirm  the change lead to the correct error message with this video: O_HLReeP_M8. I didn't try it with the one you mentioned above, but it seems that one isn't available in my country, so I might see a different message for that anyway (don't know what error takes precedence).

fent:
I get `The uploader has not made this video available in your country.` for video `O_HLReeP_M8`

Anyway, this doesn't work for me, because `info.status` is always `fail` when `info.requires_purchase` is true. Can you test if the following diff, on top of this PR, also works for you?

```
diff --git a/lib/info.js b/lib/info.js
index daa978a..72f274d 100644
--- a/lib/info.js
+++ b/lib/info.js
@@ -142,10 +142,10 @@ function gotConfig(id, options, additional, config, appendConfig, callback) {
   request(url, options.requestOptions, function(err, res, body) {
     if (err) return callback(err);
     var info = querystring.parse(body);
-    if (info.status === 'fail') {
+    if (info.requires_purchase === '1') {
+      return callback(new Error('This video requires purchase'));
+    } else if (info.status === 'fail') {
       info = config.args;
-    } else if (info.requires_purchase === '1') {
-      return callback(new Error(info.ypc_video_rental_bar_text));
     }
 
     // Split some keys by commas.
```

FD-:
Well, interestingly, info.status doesn't seem to be fail for me with the video I mentioned above. 
I get the country error when trying to access SyKPsFRP_Oc. 
The edit you suggested works fine for me.

fent:
Thanks again

FD-:
It actually seems like this change broke subtitle support. Seems like info.player_response is never reported if we request with "&el=info". I suggest to revert this change or find an alternative for getting the subtitles.

PR #207, opened 2017-08-23
Don't add additional thumbnails. (Fixes #155)

### LINKED ISSUE #155: How does one check if the video supports maxresdefault?

BrunoMoreira991:
This should fix issue [#155](https://github.com/fent/node-ytdl-core/issues/155).

Youtube API already provides us with the available thumbnails. Likewise if, for instance, `maxresdefault` isn't available the API doesn't return it. Adding them manually via the `additional` info object ends up causing this issue.

Doing something like this is possible now:
```js
var thumbnail = info.iurlmaxres || info.iurlmq;
```

### DIFF
--- a/lib/info.js
+++ b/lib/info.js
@@ -10,7 +10,6 @@ const FORMATS     = require('./formats');
 const VIDEO_URL = 'https://www.youtube.com/watch?v=';
 const EMBED_URL = 'https://www.youtube.com/embed/';
 const VIDEO_EURL = 'https://youtube.googleapis.com/v/';
-const THUMBNAIL_URL = 'https://i.ytimg.com/vi/';
 const INFO_HOST = 'www.youtube.com';
 const INFO_PATH = '/get_video_info';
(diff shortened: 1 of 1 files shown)


lib/info.js
```diff
@@ -78,13 +77,7 @@ module.exports = function getInfo(link, options, callback) {
       related_videos: util.getRelatedVideos(body),
 
       // Give the canonical link to the video.
-      video_url: url,
```
fent: Comma at end of line should not be removed.

fent:
Ah, this is true. Did not notice from the last update. Thanks!

FireController1847:
HOORAY!!!!!!!!!!!!!!!!!!!

FireController1847:
The YouTube API no longer provides these, so it might be worth re-adding them back. You're now forced to manually edit info.thumbnail_url again. UGH, YOUTUBE STOP MAKING THIS SO DIFFICULT.

PR #216, opened 2017-08-30
rename validateId to validateLink

### LINKED ISSUE #215: Need help with importing util
Hi, noob here.

I need to import util so I can use the validateLink function. The documentation is a bit unclear for me. 
In the beginning I tried to simply do `const ytdl = require('ytdl-core')` with `ytdl.validateLink(url)`, that didn't work. Then, I tried to `const {util} = require('ytdl-core')` and `util.validateLink(url)`, which didn't work either.

What is the proper way to import util?
Thanks in advance!

TimeForANinja:
atm the `validateLink` function is named
validateLink in util
validateLink in typings
validateLink in readme
and
validateId in index.js

therefor i'd just rename it in index.js to fix #215


fent:
Thanks!

Issue #226, opened 2017-09-30
Add proxy support (modify request URL before going over the network)


MadLittleMods:
Before https://github.com/fent/node-ytdl-core/commit/72395d242a7bf8398c521481c01913085c00a535 you could edit the resultant URL before it went across the network with `options.request`. It was cumbersome because of the EventEmmiter and callback pattern but was possible.

Now with `options.requestOptions`, it isn't really possible because the video URL is changed around when we fallback to the embed URL because of mature content, etc.

We could allow for `options.requestOptions = function(url) { /* ... */ }` to be a function callback passing along the URL we are eventually going to request.

---

With `ytdl-core@0.7.19` ✔️ 
```js
import Promise from 'bluebird';
import {format, parse} from 'url';
import EventEmitter from 'event-emitter';
import ytdl from 'ytdl-core';
const ytdlGetInfo = Promise.promisify(ytdl.getInfo);

const generateRequestProxy = function(opts) {
	return function(url, options, callback = () => {}) {
		const parsed = parse(url);
		const resultantOptions = Object.assign({}, parsed, options);
		const resultantUrl = `${opts.proxyUrl}${format(resultantOptions)}`;

		const requestEventEmitter = EventEmitter({});

		Promise.resolve(fetch(resultantUrl))
			.then(function(response) {
				requestEventEmitter.emit('response', response);
				return response.text();
			})
			.then(function(text) {
				callback(null, text);
				return null;
			})
			.catch(function(err) {
				callback(err);
				requestEventEmitter.emit('error', err);
			});

		return requestEventEmitter;
	};
};

const defaults = {
	proxyUrl: ''
};
function fetchYouTubeInfo(url, options) {
	const opts = Object.assign({}, defaults, options);
	return ytdlGetInfo(url, {
		request: generateRequestProxy(opts)
	});
}
```

---

With `ytdl-core@0.17.1` :x:
```js
import Promise from 'bluebird';
import {format, parse} from 'url';
import ytdl from 'ytdl-core';
const ytdlGetInfo = Promise.promisify(ytdl.getInfo);

function fetchYouTubeInfo(url, options) {
	const opts = Object.assign({}, defaults, options);
	const parsedProxy = parse(opts.proxyUrl);
	return ytdlGetInfo(url, {
                        // This doesn't work when we fallback to non-video URLs like the embed, etc.
			requestOptions: Object.assign({}, parsedProxy, {
				path: `${parsedProxy.path}${url}`
			})
		});
}
[...]

PR #227, opened 2017-09-30
Fix proxy example to actually transform URL

MadLittleMods:
Fix https://github.com/fent/node-ytdl-core/issues/226

Fix proxy example to actually transform URL.

In the example, previously `parsed` was always `null`. It looks like we actually need to use the [`miniget` library, `requestsOptions.transform` option](https://github.com/fent/node-miniget/blob/c473618fda157a2bec2fed6a4b219ae877385179/lib/index.js#L51) to get the parsed URL data passed in.


fent:
Thanks!

PR #235, opened 2017-10-15
audioencodedvideo filter added

bubundas17:
audioencodedvideo to filter formats that contain both video and audio


fent:
How about "audioandvideo" for the name of the filter? Can you also add tests? The current tests are [here](https://github.com/fent/node-ytdl-core/blob/master/test/util-test.js#L220-L246)

bubundas17:
Yap Good suggestion about naming... But, Actually I am not much familiar about this tests, can you please help me to do this?

bubundas17:
Hello! Anybody Here?

fent:
Sorry, I missed the updates to this. v_v

So for the tests, on the test file linked above, you can copy one of the blocks that test for `filter` and replace the `filter` with `audioandvideo`

```js
  describe('With `audioandvideo` given', () => {
    it('Returns only matching formats', () => {
      var itags = util.filterFormats(formats, 'audioandvideo').map(getItags);
      assert.deepEqual(itags, ['140']);
    });
  });
```

And then adjust the expected itags from `['140']` to the itags that contain both audio and video.

You can then run the `mocha test/util-test.js` command to run the tests. You can add `.only` to `describe` or `it` to run only that one test.

```js
describe.only('...', () => {});
```

bubundas17:
Ok, Done. Patched!

fent:
Thank you!

Issue #252, opened 2017-12-12
Low audio bitrate


DonikaV:
I set quality: highest, but when i download mp3, bitrate is low, 64 or 128, how can i get better quality?

fent:
The sorting function for `highest` ranks formats first by the ones that contain both video and audio, then by video resolution and bitrate, and then by audio bitrate.

https://github.com/fent/node-ytdl-core/blob/master/lib/util.js#L54

Sometimes it'll be that the formats that are ranked the highest are not the ones with the highest audio bitrate.

Usually, the format that is ranked the highest will contain the highest audio bitrate for the formats that have both audio and video. I think that's the case for every video actually.

If you want to always download the format with the highest audio bitrate regardless of it having any video, I could add a `highestaudio` option for `quality`.

However, if you want to guarantee the highest quality audio AND video, then you'd need to use something like [pully](https://github.com/JimmyBoh/pully). YouTube will often have separate formats for the highest quality of video and audio. Pully has presets to help you merge those formats into one video file with ffmpeg.

PR #267, opened 2018-01-11
add highestaudio and highestvideo

TimeForANinja:
add highestaudio and highestvideo as also kinda requested in https://github.com/fent/node-ytdl-core/issues/252
its prefering only sending one source since u got `audio` and `video` prefering both
best option for video i came up with is to compare the lowest possible bitrate, else u may have to manually add some kinda sorting index to formats.js
also added tests
was unable to run tests... istanbul error'ed out


TimeForANinja:
build is failing since first format named is
{ itag: '18',
    bitrate: '0.5',
    audioEncoding: 'aac',
    audioBitrate: 96 }
and while checking i thought this one is better...
  { itag: '43',
    bitrate: '0.5',
    audioEncoding: 'vorbis',
    audioBitrate: 128 }
not sure wether i should also add that for a 'highestvideo' option search or just change the itag

README.md
```diff
@@ -24,7 +24,7 @@ ytdl('http://www.youtube.com/watch?v=A02s8omM_hI')
 
 Attempts to download a video from the given url. Returns a readable stream. `options` can have the following keys
 
-* `quality` - Video quality to download. Can be an [itag value](http://en.wikipedia.org/wiki/YouTube#Quality_and_formats) value, a list of itag values, or `highest`/`lowest`. Defaults to `highest`.
+* `quality` - Video quality to download. Can be an [itag value](http://en.wikipedia.org/wiki/YouTube#Quality_and_formats) value, a list of itag values, or `highest`/`lowest`/`highestaudio`/`highestvideo`. `highestaudio`/`highestvideo` both prefer only responding with only audio/video. Defaults to `highest`.
```
fent: >`highestaudio`/`highestvideo` both prefer only responding with only audio/video

Not necessarily. These two are meant to return the quality with the highest audio/video regardless of the quality of the other.

If it were the case that they'd respond with only audio/video, then we could just use `quality: 'highest, filter: 'audio/videoonly'`

lib/util.js
```diff
@@ -128,6 +128,28 @@ exports.chooseFormat = (formats, options) => {
       format = formats[formats.length - 1];
       break;
 
+    case 'highestaudio':
+      var highest = formats[0];
+      for (var format of formats) {
```
fent: Use `let` for these two `var`s since these variables are redeclared in the next `case` block below

lib/util.js
```diff
@@ -128,6 +128,28 @@ exports.chooseFormat = (formats, options) => {
...
+          && (format.audioBitrate > highest.audioBitrate
+          || (format.audioBitrate === highest.audioBitrate && !format.encoding))))
+            highest = format;
+      }
+      format = highest;
+      break;
+
+    case 'highestvideo':
+      var highest = formats[0];
+      for (var format of formats) {
```
fent: Use `let` for these two also

lib/util.js
```diff
@@ -128,6 +128,28 @@ exports.chooseFormat = (formats, options) => {
       format = formats[formats.length - 1];
       break;
 
+    case 'highestaudio':
+      var highest = formats[0];
+      for (var format of formats) {
+        if (!highest.audioBitrate || (format.audioBitrate
```
fent: No need to check for `format.audioBitrate` since if it gets to this point, then we can safely assume it's truthy

lib/util.js
```diff
@@ -128,6 +128,28 @@ exports.chooseFormat = (formats, options) => {
...
+          && (format.audioBitrate > highest.audioBitrate
+          || (format.audioBitrate === highest.audioBitrate && !format.encoding))))
```
fent: No need for this line since we don't mind if there is video in this format

lib/util.js
```diff
@@ -128,6 +128,28 @@ exports.chooseFormat = (formats, options) => {
...
+      }
+      format = highest;
+      break;
+
+    case 'highestvideo':
+      var highest = formats[0];
+      for (var format of formats) {
+        if (!highest.bitrate || (format.bitrate
+          && (Number(format.bitrate.split('-')[0]) > Number(highest.bitrate.split('-')[0])
+          || (Number(format.bitrate.split('-')[0]) === Number(highest.bitrate.split('-')[0]) && !format.audioEncoding))))
```
fent: Same comments as above apply to this `if` statement. Also, the bitrate can be a float, and `util.sortFormats` looks at the right side of `bitrate`

https://github.com/fent/node-ytdl-core/blob/master/lib/util.js#L63

lib/util.js
```diff
@@ -128,6 +128,30 @@ exports.chooseFormat = (formats, options) => {
       format = formats[formats.length - 1];
       break;
 
+    case 'highestaudio':
+      formats = exports.filterFormats(formats, 'audio');
+      format = formats[0];
```
fent: Need to use `let`, otherwise, without a keyword, it's declared as a global

lib/util.js
```diff
@@ -128,6 +128,30 @@ exports.chooseFormat = (formats, options) => {
...
+      }
+      break;
+
+    case 'highestvideo':
+      function getBitrate(f) {
+        let s = highest.bitrate.split('-')
+        return parseFloat(s[s.length - 1], 10);
+      }
+      formats = exports.filterFormats(formats, 'video');
+      format = formats[0];
```
fent: Same here for `let`

lib/util.js
```diff
@@ -128,6 +128,30 @@ exports.chooseFormat = (formats, options) => {
       format = formats[formats.length - 1];
       break;
 
+    case 'highestaudio':
+      formats = exports.filterFormats(formats, 'audio');
+      format = formats[0];
+      for (let f of formats) {
+        if (f.audioBitrate > highest.audioBitrate
+          || (f.audioBitrate === highest.audioBitrate && !f.encoding))
```
fent: In the case where two formats have the same audio quality, do we prefer the format that does not contain a video? If not, then I don't think we need this line

Same for the other case below with audio/video switched

lib/util.js
TimeForANinja: format is declared above in https://github.com/TimeForANinja/node-ytdl-core/blob/master/lib/util.js#L120 to return the choosen format and formats is declared as param passed to the func
i would actually just overwrite those

lib/util.js
TimeForANinja: the problem i'm having is that most ppl using `highestaudio` and `highestvideo` will most likely merge them or u'd pick them 'cause ur not interested in the other and most likely removing it from the stream anyway
the other problem i'm personaly having is that using both `highestaudio` and a filter `audioonly` results in an error getting throw since some videos actually only support one format(for my e.g. itag 18)

fent: >the problem i'm having is that most ppl using highestaudio and highestvideo will most likely merge them or u'd pick them 'cause ur not interested in the other and most likely removing it from the stream anyway

Hmm, in that case I'd change the above description in the readme to

    `highestaudio`/`highestvideo` both prefer audio/video only respectively.


>the other problem i'm personaly having is that using both highestaudio and a filter audioonly results in an error getting throw since some videos actually only support one format(for my e.g. itag 18)

What's the error getting thrown? We should probably fix that, doesn't need to be in this PR though

TimeForANinja: >> the other problem i'm personaly having is that using both highestaudio and a filter audioonly results in an error getting throw since some videos actually only support one format(for my e.g. itag 18)
>
>What's the error getting thrown? We should probably fix that, doesn't need to be in this PR though

did a typo there, was actually using `{filter: 'audioonly'}` which ended with `new Error('No formats found with custom filter');` ([L116](https://github.com/fent/node-ytdl-core/blob/master/lib/util.js#L116))
so no need to fix that...

lib/util.js
```diff
@@ -128,6 +128,30 @@ exports.chooseFormat = (formats, options) => {
...
+      function getBitrate(f) {
+        let s = f.bitrate.split('-');
+        return parseFloat(s[s.length - 1], 10);
+      }
+      formats = exports.filterFormats(formats, 'video');
+      format = formats[0];
+      for (let f of formats) {
+        if (getBitrate(f) > getBitrate(format)
+          || (getBitrate(f) === getBitrate(format) && format.audioEncoding && !f.audioEncoding))
+            format = f;
```
fent: This line is currently not being test covered: https://codecov.io/gh/fent/node-ytdl-core/compare/692da6cc6d7206f8149382186742fa98299eb2ec...9fbcf14693ae97c03a327596f7ab191cb1dc587f/diff

Can be fixed by either
* setting the above `format`s to `format = null;` and then adding another check in the loops for `!format`
* changing the order of `formats` in `util-test.js` so that the format at the top of the list is not the one we initially want, by calling `.reverse()` on it for example

fent:
Awesome. Thanks for the work! :)

PR #289, opened 2018-02-09
Problem with key valid returning error

Santimto:


fent:
Could you post a video where the `key` isn't found?

Santimto:
https://www.youtube.com/watch?v=_OPTyIrhwX8
https://www.youtube.com/watch?v=pAzaqRpmAGc
https://www.youtube.com/watch?v=bAu6d0r50iY

fent:
I'm still able to parse a `key` with all of those videos. And I don't get the point of negating the `key` check, if you look at the travis-ci builds, the tests fail.

But I do think the change has a point, we don't have to check if `key` is defined or not.

Santimto:
I'm using in my company's application with this change and stopped giving errors ... without it's common error: Could not extract signature deciphering actions that force restart of the application.

fent:
Are you sure fixing that error isn't from a recent release that fixes it?

https://github.com/fent/node-ytdl-core/commit/18b3f22ca14b6ee61bc646fd8a2f2c943baa6709

Santimto:
Fixed in last update! Thanks =)

PR #290, opened 2018-02-14
Move getBitrate func to fix crash in react native

khaled-cliqz:
React Native engine doesn't allow declaring functions in nested lexical scopes e.g. switches.
it throws `Strict mode does not allow function declarations in a lexically nested statement` error


fent:
Didn't know this, will keep in mind. Thank you!

PR #338, opened 2018-07-12
Possible fix for ECONNRESET

### LINKED ISSUE #331: discord.js music bot using ytdl having playback problem
I am making a discord music bot using the ytdl-core library. I also use node-opus and ffmpeg. So, the dependencies are not an issue. 
The bot loads the song fine, its the playback thats having problem. Here is the hastebin for my code: https://hastebin.com/dokoguqoqu.js

It successfully outputs Now Playing which means there are no errors. However right after the bot joins the voice channel it never plays the song instead a second after it joins it executes the dispatcher.on(end) command and outputs Finished Playing. 

The audio from the youtube video is loaded but never played.

### LINKED ISSUE #332: ECONNRESET
It's baack!

My bot is almost always returning an ECONNRESET with its streams. This was caught after using `stream.on("error")` and an extremely long wait time before data is emitted.

FireController1847:
Fixes #331 and Fixes #332

I noticed that honestly the biggest difference between downloadFromInfo and the main function is that downloadFromInfo uses `immediate` while the main function does not. I edited the main function to use setImmediate and I have yet to experience an ECONNRESET. Possible fix? It'd be awesome if I could get some more users to test this.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -19,8 +19,9 @@ const ytdl = module.exports = function ytdl(link, options) {
       stream.emit('error', err);
       return;
     }
-
-    downloadFromInfoCallback(stream, info, options);
+    setImmediate(() => {
+      downloadFromInfoCallback(stream, info, options);
+    });
   });
 
   return stream;


fent:
After some time, does this still work for you? And is your code like on #332 ?

FireController1847:
I've been using my development bot for the testing of this. Sadly I can't give you a solid answer because I've had unrelated issues on my development bot. I would use it on my production bot, but I already fetch the info beforehand and it's always been that way so I can't give you a solid yes or no. From the amount of time I have tested with the development bot, though, it seems to work great. Not one ECONNRESET.

Here's a link to the line of code that works on my new bot. [visualfiredevelopment/fm-radio#L242](https://gitlab.com/visualfiredevelopment/fm-radio/blob/v6.2/src/Modules/Player.js#L242)

FireController1847:
I think this is a working fix @fent.

And, logically, that'd make sense. The only difference between downloadFromInfo and the main function honestly is the fact the info is gotten a couple milliseconds beforehand (and when I run the two functions right next to each other aka doing exactly what download does, it causes eocnnreset while with setImmediate it does not).

FireController1847:
Closed ref https://github.com/fent/node-ytdl-core/issues/332#issuecomment-407950341

PR #340, opened 2018-07-22
[fix] add an extra check in getAuthor from utils, so that error isn't…

lokiiarora:
… stagnant and we can check for it when using the library


kelyvin:
+1 

This pull request will fix the issue that I am encountering, which I have documented here: https://github.com/fent/node-ytdl-core/issues/349

PR #342, opened 2018-07-26
Add Author Verification

FireController1847:
Is the author of this video verified? Let's find out! One simple regex and a one line if statement and we're on our way.

Oh, and I fixed one regexp inconsistency because it was annoying me lol (one of them was RegExp and one was Regexp so I standardized it with Regexp)

Btw here's some test videos:

```js
const vid = "https://www.youtube.com/watch?v=AWGqoCNbsvM"; // Verified ~ True
const vid = "https://www.youtube.com/watch?v=PMivT7MJ41M"; // Official Artist ~ False
const vid = "https://www.youtube.com/watch?v=RQGa0DPwes0"; // Not Verified ~ False
```

Example output:
```json
{
   "author":{
      "id":"UCBlXovStrlQkVA2xJEROUNg",
      "name":"Marcel Vos",
      "avatar":"https://yt3.ggpht.com/a-/ACSszfHRXQKxqrx4usYP5fQxjRcwUU2f-sMIC5exJQ=s88-mo-c-c0xffffffff-rj-k-no",
      "verified":true,
      "user":null,
      "channel_url":"https://www.youtube.com/channel/UCBlXovStrlQkVA2xJEROUNg",
      "user_url":null
   }
}
```

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -395,7 +395,7 @@ exports.getAuthor = (body) => {
     name: channelMatch[2],
     avatar: url.resolve(VIDEO_URL, exports.between(ownerinfo,
       'data-thumb="', '"')),
-    verified: verifiedMatch && !!verifiedMatch[1],
+    verified: !!verifiedMatch && !!verifiedMatch[1],
     user: userMatch ? userMatch[1] : null,
     channel_url: 'https://www.youtube.com/channel/' + channelMatch[1],
     user_url: userMatch ? 'https://www.youtube.com/user/' + userMatch[1] : null,
--- a/lib/util.js
+++ b/lib/util.js
@@ -395,7 +395,7 @@ exports.getAuthor = (body) => {
     name: channelMatch[2],
     avatar: url.resolve(VIDEO_URL, exports.between(ownerinfo,
       'data-thumb="', '"')),
-    verified: verifiedMatch != null && verifiedMatch[1] != null ? true : false,
+    verified: verifiedMatch && verifiedMatch[1] ? true : false,
     user: userMatch ? userMatch[1] : null,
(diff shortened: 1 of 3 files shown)


lib/util.js
```diff
@@ -387,12 +388,14 @@ exports.getAuthor = (body) => {
...
   const channelMatch = ownerinfo.match(authorRegexp);
-  const userMatch = ownerinfo.match(aliasRegExp);
+  const userMatch = ownerinfo.match(aliasRegexp);
+  const verifiedMatch = ownerinfo.match(verifiedRegexp);
   return {
     id: channelMatch[1],
     name: channelMatch[2],
     avatar: url.resolve(VIDEO_URL, exports.between(ownerinfo,
       'data-thumb="', '"')),
+    verified: verifiedMatch && verifiedMatch[1] ? true : false,
```
fent: Since this ternary expression gives `true/false`, we can do `verifiedMatch && !!verifiedMatch[1]`

FireController1847: Wow, I've never used something like that (a couple exclamation mark), I'll make that change now.

FireController1847:
Um, I think there's an issue with your suggestion. On the fact that the match doesn't exist (the user is not verified), it will return null due to the first statement not working, hence it returning `null` not `false`.

fent:
> Um, I think there's an issue with your suggestion. If the first match doesn't exist (aka the user is not verified), it will return null not false.

Oh right, sorry. In that case, we'll have to cast both variables, or the result of both, to a bool.

```js
!!verifiedMatch && !!verifiedMatch[1]
```

or

```js
!!(verifiedMatch && verifiedMatch[1])
```

fent:
Thanks!

PR #343, opened 2018-07-26
Add Video Extras

### LINKED ISSUE #329: Is it possible to add "video extras" to video info?
I've been attempting this recently, but in the end have failed.

What I need is the "extra" info found under the video description, including the category, license, and if possibly the "music in this video". I cannot seem to find a way to do this regex. The image below is what I'm talking about.
image
I can see this being extremely useful in applications and particularly useful to what I'm attempting to do.

FireController1847:
Fixes #329

Fetches a YouTube's video extra data which can be found under the "learn more".

Example output:
```json
{
   "extras":{
      "category":"Music",
      "song":"That's What I Like",
      "artist":"Bruno Mars",
      "writers":"Bruno Mars, James Fauntleroy, Jonathan Yip, Brody Brown, Philip Lawrence, Ray McCullough, Jeremy Reeves, Ray Romulus",
      "licensed_by":"WMG (on behalf of Atlantic Records); Warner Chappell, UBEM, UMPG Publishing, SOLAR Music Rights Management, Sony ATV Publishing, PEDL, UMPI, BMG Rights Management, ARESA, ASCAP, and 13 Music Rights Societies"
   }
}
```

Also updates the "related video" for typings. The "related video" object for the tests is outdated, so there technically were to "video extras" for it. All I've done is re-requested the same video and updated it.


lib/util.js
```diff
@@ -369,6 +369,32 @@ exports.getVideoDescription = (html) => {
     new Entities().decode(exports.stripHTML(description[1])) : '';
 };
 
+/**
+ * Get video extras from html
+ * 
+ * @param {String} body
+ * @return {Object}
+ */
+const extrasRegexp = /<h4 class="title">[\s\S]*?([\s\S]*?)<\/h4>[\s\S]*?<ul.*?class=".*?watch-info-tag-list\">[\s\S]*?<li>([\s\S]*?)<\/li>/g;
```
fent: Using `.exec()` below changes the state of this regexp. Define it inside `getVideoExtras()` so that a fresh state is used every time. Otherwise, it will only find extras for the first video.

lib/util.js
```diff
@@ -369,6 +369,32 @@ exports.getVideoDescription = (html) => {
...
+    '<div id="watch-discussion" class="branded-page-box yt-card">');
+  if (extrainfo === '') {
+    return {};
+  }
+  extrainfo = new Entities().decode(extrainfo);
+
+  const contentRegexp = /<a.*?>(.*?)<\/a>/;
+  const extras = {};
+  let match = extrasRegexp.exec(extrainfo,toString());
+  while (match) {
```
fent: Can do `while ((let match = extrasRegexp.exec(extrainfo)) != null) {` to save writing `.exec` twice in two lines

lib/util.js
fent: Regexp doesn't need the first `[\s\S]*?`, `"` doesn't need to be escaped

lib/info.js
```diff
@@ -75,6 +75,9 @@ module.exports = function getInfo(link, options, callback) {
       // Get description.
       description: util.getVideoDescription(body),
 
+      // Get extra info.
+      extras: util.getVideoExtras(body),
```
fent: Hmm, not completely sure about the name `extras`. How about `media`, it's always a song, game, or movie right?

FireController1847: Tbh the only reason I called it `extras` is because of the YouTube Tag itself being called `<div id="watch-description-extras">`. But sure, I'll rename it to media. Makes more sense tbh, although I've seen a "License" field in some videos but I have yet to test it on one of those but I guess that could be considered media aswell lol.

lib/util.js
FireController1847: This isn't working, unexpected token at `let`

FireController1847:
Odd build error, I don't think it's a code issue more a Travis CI issue 🤔

lib/util.js
```diff
@@ -369,6 +369,30 @@ exports.getVideoDescription = (html) => {
...
+exports.getVideoMedia = (body) => {
+  let mediainfo = exports.between(body,
+    '<div id="watch-description-extras">',
+    '<div id="watch-discussion" class="branded-page-box yt-card">');
+  if (mediainfo === '') {
+    return {};
+  }
+  mediainfo = new Entities().decode(mediainfo);
+
+  const regexp = /<h4 class="title">([\s\S]*?)<\/h4>[\s\S]*?<ul.*?class=".*?watch-info-tag-list\">[\s\S]*?<li>([\s\S]*?)<\/li>/g;
```
fent: No need to escape the `"` character in `watch-info-tag-list\"`

lib/util.js
```diff
@@ -369,6 +369,30 @@ exports.getVideoDescription = (html) => {
...
+  const contentRegexp = /<a.*?>(.*?)<\/a>/;
+  const media = {};
+  while ((match = regexp.exec(mediainfo)) != null) {
```
fent: `match` is not defined. My suggestion from before was not correct sorry. Syntax is more like

```js
let match;
while ((match = ...) != null) {
```

lib/util.js
FireController1847: Oops, sorry

lib/util.js
FireController1847: Ah yes, sorry. Never seen this syntax before, don't use while loops very often lol

lib/util.js
```diff
@@ -369,6 +369,31 @@ exports.getVideoDescription = (html) => {
...
+ * @return {Object}
+ */
+exports.getVideoMedia = (body) => {
+  let mediainfo = exports.between(body,
+    '<div id="watch-description-extras">',
+    '<div id="watch-discussion" class="branded-page-box yt-card">');
+  if (mediainfo === '') {
+    return {};
+  }
+  mediainfo = new Entities().decode(mediainfo);
```
fent: This should be done after all the regexp searching and matching, for `match[2]`, and maybe for `match[1]`.

Otherwise, the content of a tag could be something like

    <li>Faded (Lost Stories Remix)&lt;/li&gt; this part is ignored although it shouldn't be</li>

lib/util.js
```diff
@@ -369,6 +369,31 @@ exports.getVideoDescription = (html) => {
...
+
+  const regexp = /<h4 class="title">([\s\S]*?)<\/h4>[\s\S]*?<ul.*?class=".*?watch-info-tag-list">[\s\S]*?<li>([\s\S]*?)<\/li>/g;
+  const contentRegexp = /<a.*?>(.*?)<\/a>/;
+  const media = {};
+  let match;
+  while ((match = regexp.exec(mediainfo)) != null) {
+    if (match[2].match(contentRegexp)) match[2] = contentRegexp.exec(match[2])[1];
```
fent: Let's avoid running `.match` and `.exec` more than needed.

```js
let linkContent = contentRegexp.exec(match[2])
if (linkContent) match[2] = linkContent[1]
```

or

```js
let linkContent = contentRegexp.exec(match[2])
... = linkContent ? linkContent[1] : match[2];
```

fent:
Thank you for all the PRs lately!

FireController1847:
No problem! Lol, I was just bored recently and thought it'd be a good challenge to learn that Regexp and how this all works :p

PR #344, opened 2018-07-26
Update Test "Related Videos"

FireController1847:
This must be merged first for #343 to work.

The "related videos" object for the tests is outdated, so there technically were to "video extras" for it. All I've done is re-requested the same video and updated it.


FireController1847:
Turns out this wasn't needed after I fixed the regex. >.<

Issue #349, opened 2018-08-02
Exception thrown with .getAuthor and failed regex


kelyvin:
Hi, 

I sometimes run into the following exception that is thrown when calling `.getInfo`. 

```
'TypeError: Cannot read property \'1\' of null',
     '    at Object.exports.getAuthor (/Users/Kelvin/Workspaces/app/node_modules/ytdl-core/lib/util.js:419:21)',
     '    at request (/Users/Kelvin/Workspaces/app/node_modules/ytdl-core/lib/info.js:52:20)',
     '    at IncomingMessage.res.on (/Users/Kelvin/Workspaces/app/node_modules/miniget/lib/index.js:76:11)',
     '    at emitNone (events.js:111:20)',
     '    at IncomingMessage.emit (events.js:208:7)',
     '    at endReadableNT (_stream_readable.js:1055:12)',
     '    at _combinedTickCallback (internal/process/next_tick.js:138:11)',
     '    at process._tickCallback (internal/process/next_tick.js:180:9)
```

This occurs in the following line https://github.com/fent/node-ytdl-core/blob/master/lib/util.js#L415 because `channelMatch` results in a `null` value. 

This seems to occur sporatically when fetching "author" data. I'm not sure if a potential better solution exists, but a quick non-breaking solution is to create a check to see if the regex returns a proper result, and if it doesn't, simply return an empty object for `author`. In its current state, the request is async so I cannot properly handle the exception when this occurs either. :(

Thanks for the help!

kelyvin:
Looks like this is potentially solved with the following pull request: https://github.com/fent/node-ytdl-core/pull/340

fent:
I should have asked what videos this happens on? Better solution is to look at those videos and see where the author info is on those pages.

fent:
Will keep this opened until we find those videos and parse the author info properly.

kelyvin:
It occurred on the first youtube link that I tried. The one I received the exception on was:
https://www.youtube.com/watch?v=elwTgpHlty0

It happens sporadically (i.e. sometimes the author metadata was returned, and sometimes it didn't). Didn't get a chance to see the payload that was being parsed.

fent:
Are you passing any options to ytdl?

kelyvin:
Nope no options. My code block was simply:

```javascript
return ytdl.getInfo(videoId)
    .then((info) => {
     ...
    })
```

fent:
[...]

PR #350, opened 2018-08-03
[Fix] Check null value for channelMatch within getAuthor from utils

kelyvin:
To address issue: https://github.com/fent/node-ytdl-core/issues/349


fent:
Thanks!

PR #378, opened 2018-11-09
readme correction

Luxray5474:
or it could just be the value of the itag value value, or maybe the value of said value, or maybe the itag value of the itag value of the value of the itag value.


fent:
Thanks!

PR #381, opened 2018-11-27
Move Readme "Tags"

FireController1847:
Move readme tags to the right of the title like every other library lol

(imho it looks easier to read and cuts off the text less)


fent:
I used to put the badges next to the module name long ago. But I moved them to after the short description, mainly because I thought the description was more important to immediately see when scrolling down, compared to the badges.

It's also easier to edit badges if they're on separate lines. Might still be able to do this and at the same time make them show up next to the name, not sure.

FireController1847:
I see where you're going and I totally understand, although I think in their current position the badges are more distracting than they would be as if they were next to the title. For a couple reasons, the main one being that since everyone is already used to it being next to the title they tend to get ignored. Right now, they actually _cut off_ the description abruptly, and while I was reading I lost my place and was discouraged to continue reading (hence the creation of this PR).

fent:
Hmm, ok. How about placing them right below the name, each on their own line for easier editing.

FireController1847:
I tried that but for some reason it put them all on separate lines

fent:
How about like this? They're on separate lines in the file, but show up on the same line rendered.

FireController1847:
Looks good to me. It doesn't cut off the description and is still very smooth to read.

fent:
Great, and thanks again!

PR #396, opened 2019-01-06
add lowestaudio / lowestvideo

TimeForANinja:
respond to https://github.com/fent/node-ytdl-core/issues/395

PR #397, opened 2019-01-08
Fixed type declarations

Favna:
This was skimmed over for v0.29.0, and actually some X earlier release as well since it was also missing `highestaudio` and `highestvideo`. 

I haven't really looked into the rest to be honest so maybe there is more but for now at least this addresses #396 

Adding typings helps for IDE support, for example in IntelliJ / VSCode IntelliSense can suggest the proper values:

image

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -4,7 +4,7 @@ declare module 'ytdl-core' {
 
   namespace ytdl {
     type downloadOptions = {
-      quality?: 'lowest' | 'highest' | string | number;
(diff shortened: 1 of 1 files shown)


Favna:
Travis needs to get a kick on the ass. That fail is not at all related to the PR smh >:C

fent:
Thank you!

Favna:
@fent would you mind releasing the change to v0.29.1 on NPM so I can use a version that npmignore into account?

fent:
Done

PR #401, opened 2019-01-17
README cleanup, yarn to gitignore & templates to .github folder

Favna:
- I moved the issue template to `.github` folder to get a cleaner root folder (ye this works)
- I created an `.npmignore` folder. I see that npm builds already have files filtered out but since you didn't have this I can only imagine you've been manually creating the tarball or pulling files out. This is the new auto generated tarball with the file:

```nginx
$ npm publish --dry-run
npm notice
npm notice package: ytdl-core@0.29.2
npm notice === Tarball Contents ===
npm notice 1.2kB  package.json
npm notice 1.1kB  LICENSE.md
npm notice 8.5kB  README.md
npm notice 13.2kB lib/formats.js
npm notice 4.2kB  lib/index.js
npm notice 3.9kB  lib/info-extras.js
npm notice 11.1kB lib/info.js
npm notice 7.4kB  lib/sig.js
npm notice 9.7kB  lib/util.js
npm notice 7.8kB  typings/index.d.ts
npm notice === Tarball Details ===
npm notice name:          ytdl-core
npm notice version:       0.29.2
npm notice package size:  18.3 kB
npm notice unpacked size: 68.3 kB
npm notice shasum:        44b5f09a8834d747ae6b26d21ebdb81b99782f49
npm notice integrity:     sha512-FpC7XUB6R9tdy[...]Bd+XRfDV58Wow==
npm notice total files:   10
npm notice
+ ytdl-core@0.29.2
```

- Markdown-ified the License file because markdown is bae
- Added some typescript and yarn related documentation to the README
- Added `yarn.lock` to gitignore so contributors can use yarn without having to remove the file less it gets committed

I've split all these changes in a bunch of tiny commits so each can be easily reverted if requested.


Favna:
Travis failed for the same reasons on 0496ebca1e800069af34f43ad502925715e8f6e2 👀

.npmignore
```diff
@@ -0,0 +1,13 @@
+node_modules
```
fent: Since `package.json` includes a list of `files`, this isn't needed. `files` tells npm what to include when publishing. See more at https://docs.npmjs.com/files/package.json#files

LICENSE.md
```diff
@@ -0,0 +1,25 @@
+The MIT License (MIT)
```
fent: Don't think this is necessary. The license file is currently simple enough to be in plain text.

README.md
```diff
@@ -16,6 +16,13 @@ You can contact us for support on our [chat server](https://discord.gg/V3vSCs7)
 ```js
 const fs = require('fs');
+/**
+ * or for Babel / TypeScript:
+ * import ytdl from 'ytdl-core';
+ * Note: Requires --esmoduleinterop for TypeScript
+ * use "import * as ytdl" with --allowsyntheticdefaultimport
+ * or "import ytdl = require('ytdl-core')" with neither
```
fent: Maybe this line should go at the top since it uses no options? And the current first one format would look better with the format from the current 2nd. So something like

    import ytdl from 'ytdl-core'; with --esmoduleinterop
    import * as ytdl from 'ytdl-core'; with --allowsynthethicdefaultimport

README.md
```diff
@@ -155,6 +162,8 @@ For getting started with that, you can look at the `extractActions()` function i
 
 ```bash
 npm install ytdl-core
+
+# Replace "npm install" with "yarn add" when using yarn
```
fent: Would prefer subsections for both npm and yarn, instead of a comment

Favna: The only difference is that single substitution though, the rest is 100% the same. Should I just add a new code block then?

Favna:
@fent addressed some of your feedback, I had some questions about the 2 bits about README, posted the questions in the Discord server

README.md
```diff
@@ -16,6 +16,9 @@ You can contact us for support on our [chat server](https://discord.gg/V3vSCs7)
 ```js
 const fs = require('fs');
+// TypeScript: import ytdl from 'ytdl-core'; with --esModuleInterop
+// TypeScript: import * as ytdl from 'ytdl-core'; with --allowSyntheticDefaultImports
+// TypeScript: import ytdl = require('ytdl-core'); with neither of the above
```
FireController1847: I'm not entirely sure why we would include every example of TypeScript. Surely if they were using TypeScript they'd know how to import a module no?

README.md
```diff
@@ -157,6 +160,11 @@ For getting started with that, you can look at the `extractActions()` function i
 npm install ytdl-core
 ```
 
+Or for Yarn users:
+```bash
+yarn add ytdl-core
+```
+
```
FireController1847: I think instead of adding/changing this, we should just remove it outright. Like with TypeScript, surely anyone using Yarn would know how to add a node module.

Favna:
@FireController1847 I'm going to reply to both of your comments in 1 go from here. 

I do not necessarily disagree with your point of view, however I also do not see a reason not to list these things. While for TypeScript you have a good point that it is more than likely that TypeScript users would be more advanced in their knowledge I have to say that when I personally adopted TypeScript after more than a year of being in JavaScript based NodeJS I did have some big struggles of getting everything set up the way I wanted it to and learning how module loading works. A little hint here and there wouldn't have hurt I'd say.

As for yarn, considering how big it is I wouldn't be surprised if there are people out there who follow a starter tutorial that make them use Yarn from the get-go. 

Lastly lets not forget: "never assume, you only make an ass out of u and me". That is to say, it's bad to assume and better to cover your bases.

All in all, like I said I do not necessarily disagree. Lets wait for @fent to share his thoughts since it's his project at the end of the day.

fent:
@FireController1847 I didn't see your comment before merging. Had this opened for a while in a tab and just now merged without checking. :(

Sorry. About your comments. I somewhat agree that after a while, instructions on how to import or install a library get mundane if there are various solutions out there. Ideally, we wouldn't put these instructions in the readme at all, only that it's a node module, and its usage instructions. And then the user would know how to install and import the package based on its name or github url.

Maybe it could be a setting for the repo. Like how the license type is labeled at the top of a repo now, github didn't use to have that, and packages were more likely to list their license in the readme. If you look through the commit history, I removed the license from the readme around the time that happened, because a cleaner readme is nicer :)

But, such setting for package type doesn't exist yet, so we have to tell people that this is usually installed for node through npm. I agree that if someone is using yarn, they're more likely to know the default package manager for node is npm, and to replace any instances of "npm install" with "yarn add" in instructions. But I also don't mind listing it. I won't go out of my way to, but I don't say no if someone does it. :) I wouldn't mind instructions on how to use this in the browser or in react-native either, since we've gotten a few issues asking those questions before.

Anyway, if we're not quite sure this is the format to go with in the readme, we can always make further changes.

FireController1847:
Alright, that's perfectly fine. Just wanted to give my opinion :)

PR #404, opened 2019-01-19
Update LICENSE

FireController1847:
This kind of bothered me -- let's ensure your work is fully "protected".

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (C) 2012 by fent
+Copyright (C) 2019 by fent
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (C) 2019 by fent
+Copyright (C) 2012-2019 by fent
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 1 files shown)


FireController1847:
If I knew this was going to be PR #404, I would have made it just for that reason lol

fent:
Is this a requirement for licenses? I don't wanna have to update it every year. Even if it's just one commit a year, I have several libraries I'm maintaining. I'm imagining the commit clutter once a library is in "maintenance" mode. Is there another change we could make, like putting `present` instead of `2019`?

I looked through a few popular node modules, which I found on the npmjs.com homepage

no year:
- https://github.com/browserify/browserify/blob/master/LICENSE
- https://github.com/gruntjs/grunt/blob/master/LICENSE
- https://github.com/npm/cli/blob/latest/LICENSE
- https://github.com/apache/cordova-cli/blob/master/LICENSE

one year, but outdated:
- https://github.com/foreverjs/forever/blob/master/LICENSE

has a range of years, but outdated:
- https://github.com/gulpjs/gulp/blob/master/LICENSE

has present for the latter year:
- https://github.com/bower/bower/blob/master/LICENSE

several ranges of years across different contributors
- https://github.com/expressjs/express/blob/master/LICENSE

FireController1847:
As far as I know, the "LICENSE" technically has no legal binding. However, _I am not a lawyer_, so don't take my word for any of this lol. But I'll update it to say "present," which is probably a better idea.

PR #414, opened 2019-02-16
Added getVideoURLwithID 

ThatZiv:
Returns video link by only using the video id as the parameter.


FireController1847:
Is this really a necessary addition? Couldn't this just be concatenated within the user's code?

fent:
Not sure this is necessary. What's your user case?

ThatZiv:
Sorry, I'll close it

Issue #412, opened 2019-02-10
Access video URL from another machine and IP address


EddieEJ:
When calling `getInfo` method and obtaining all the links from formats, on some videos(not all) I do not have access to the direct video link. When trying to load the link into my browser from a different device with different network and ip address, I simply get:
![image](https://user-images.githubusercontent.com/25306489/52528648-56f70500-2cdb-11e9-8dbc-c10b3f55346f.png)

How can I allow my users to retrieve video URLs which can be playable on their machines without having to download the video?

WaqasIbrahim:
Facing the same issue. Some links also return 404.
https://www.youtube.com/watch?v=Mt6CCmUHfm8 returns a 404

EddieEJ:
So after looking at the [rg3/youtube-dl README](https://github.com/rg3/youtube-dl), more specifically [I extracted a video URL with -g, but it does not play on another machine / in my web browser](https://github.com/rg3/youtube-dl#i-extracted-a-video-url-with--g-but-it-does-not-play-on-another-machine--in-my-web-browser) I guess this is a common issue. 

However, since I do not wish to download the videos onto my users' devices. What might be a workaround for this? How can I use the url obtained and be able to play it on different machines with different IP addresses? Also, does the url ever expire? 

Any help would be appreciated.

fent:
As far as I know, download URLs are tied to a specific IP. And thus you cannot share a download URL with others. You can have noticed that if you have a long video opened, and then switch networks, or use a proxy, youtube will reload the video. It doesn't always happen, did you notice what type of videos were the ones where this worked or didn't work on?

> However, since I do not wish to download the videos onto my users' devices. What might be a workaround for this?

Are you providing the download link without the intention of the users downloading it?

fent:
> Also, does the url ever expire?

Just checked, 6 hours. The download URLs contain an `expire` key with a unix timestamp. I remember this being 4 hours some time ago, so if you want to be sure, check this key.

I'm going to keep this issue opened as a reminder to add this to the readme, since it's frequently asked.

sesam:
> I'm going to keep this issue opened as a reminder to add this to the readme, since it's frequently asked.

See https://github.com/sesam/node-ytdl-core/commit/6e2524b9fda10e5afffb023674f9f5919622ff08

Enough? Is this issue done?

baptisteArno:
So there is no work around for this? :'(

[...]

PR #423, opened 2019-04-18
Inform on timing and IP-address link

sesam:
Adresses #412

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,8 @@ Same as the above `ytdl.getURLVideoID()`, but can be called with the video ID di
 
 ## Limitations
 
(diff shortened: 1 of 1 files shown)


fent:
Thanks! Can you move it to the end of the "Limitations" section? And also remove "computer" and just put "ip address" in there?

sesam:
Sure. I think I'll also have to add a qualifier, like "possibly" for the IP lock. I just made a test that uses ytdl-core in AWS lambda, then gets access to our test mp4 video via a Heroku free account, which at least usually would have a different first IP octet.

fent:
hi, any new information you've found on this?

PR #428, opened 2019-05-10
Fix 403 forbidden error

### LINKED ISSUE #417: Very rare 403 error when streaming from YouTube
Very rarely I get a 403 error when trying to stream a YouTube video with my discord bot.
However, I can then immediately retry to play the exact same video and then it works.

 I haven't found a consistent way of reproducing this error and it is quite rare, maybe 1 in 1000 stream attempts would be my guess. Also continuously trying to stream the same video doesn't seem to be able to reproduce the error. So any attempt would have to probably use a collection of a lot of different videos (1000+) and have a lot of time, but hopefully there is another way of figuring this out.

Here is the stack trace (which seems quite useless, as it's just the status code that miniget receives):

```
Error: Status code: 403
  at ClientRequest.httpLib.get (/REDACTED/node_modules/miniget/lib/index.js:123:19)
  at Object.onceWrapper (events.js:315:30)
  at emitOne (events.js:116:13)
  at ClientRequest.emit (events.js:211:7)
  at HTTPParser.parserOnIncomingClient [as onIncoming] (_http_client.js:543:21)
  at HTTPParser.parserOnHeadersComplete (_http_common.js:112:17)
  at TLSSocket.socketOnData (_http_client.js:440:20)
  at emitOne (events.js:116:13)
  at TLSSocket.emit (events.js:211:7)
  at addChunk (_stream_readable.js:263:12) +0ms
```

The error always seems to appear twice (I only posted one to keep the text shorter), completely identically. I thought your miniget module would only retry on a 500 error, so I'm not sure why it shows up twice.

I am using the newest ytdl-core like this `stream = ytdl(url, opts);`.
And I'm using discord.js on the master branch, where I pass the stream like this `this.conn.play(this.stream, opts);`

I think if ytdl-core could simply retry once after receiving a 403, this might solve the problem. Of course a 403 with `This video is not available.` would need no retry. If I'm not mistaken ytdl-core doesn't retry on a receiving a 403 as of now, of course rightly so as normally a 403 means that the video is not available.

In this case, I think something might be going wrong on YouTube's side, but we can't just wait for them to fix it. And I know that this is quite a difficult problem, as it is not easily reproducible. 

If you do not want to change anything in this module, because this error is very vague, I would gladly try out a changed version myself, I might just need a hint how I could add a retry when a 403 is received.

BrunoMoreira99:
Fixes #417.

This error has been happening to me way too frequently, after I saw https://github.com/fent/node-ytdl-core/issues/417#issuecomment-468048467 from @omarroth I was able to fix this issue. So, thanks 👍 I wouldn't have had a clue how to fix this otherwise.

I tested this thoroughly and was able to confirm that this does indeed fix the 403 forbidden error.

I went ahead and added some comments also, please tell me if you'd prefer no comments.
There is a much more detailed explanation of the issue at https://github.com/ytdl-org/youtube-dl/pull/18927.

### DIFF
--- a/lib/sig.js
+++ b/lib/sig.js
@@ -246,9 +246,9 @@ exports.setDownloadURL = (format, sig, debug) => {
     // into the parameter it specifies.
     // See https://github.com/fent/node-ytdl-core/issues/417
     if (format.sp) {
-        query[format.sp] = sig;
+      query[format.sp] = sig;
     } else {
-        query.signature = sig;
+      query.signature = sig;
     }
   }
 
--- a/lib/sig.js
+++ b/lib/sig.js
@@ -242,7 +242,14 @@ exports.setDownloadURL = (format, sig, debug) => {
   // See https://github.com/fent/node-ytdl-core/issues/127
   query.ratebypass = 'yes';
   if (sig) {
-    query.signature = sig;
+    // When YouTube provides a `sp` parameter the signature `sig` must go
+    // into the parameter it specifies.
+    // See https://github.com/fent/node-ytdl-core/issues/417
+    if (format.sp) {
+        query[format.sp] = sig;
+    } else {
+        query.signature = sig;
+    }
   }
 
   format.url = url.format(parsedUrl);


BrunoMoreira99:
On a side note, I don't think this function is actually doing anything right now. https://github.com/fent/node-ytdl-core/blob/43c045ede867d8fac80e95f56bb08f1057866893/lib/info.js#L246-L250
I tackled around that function a bit because I thought I'd need to change it as well, but I came to the conclusion that actually the URLs it is meant to decipher (which I assume are only those from livestreams, usually) don't actually have the signature after `s` but after `sig` or `signature`, and I think that signature is already deciphered (as I tried to decipher it but THEN it would throw a 403 error), in fact, I tried not using this function at all and just passing the url directly to `getDashManifest` and `getM3U8` and it worked, although the livestream actually only plays for a few seconds, but I believe that is a whole different issue.

TheAkio:
The issue with the stream stopping after a few seconds might be related to YouTube having a very weird way of doing MPEG-TS Livestreams. I made a small library that basically deals with that stuff before piping it to FFMPEG but it is also broken right now. I wanted to check over the next few days what's causing the issue. Might just be some small changes since I made that library. You can check it out here: https://github.com/TheAkio/ytls

I'm not actively working on that library and I don't plan to, so if you wanted we could build that handling into ytdl-core directly.

fent:
Thank you for this.

Still in the middle of fixing tests, but merging this in and releasing a patch version.

PR #481, opened 2019-08-30
Fixed related videos returning empty array

### LINKED ISSUE #479: related_videos is always empty
[Issuehunt badges](https://issuehunt.io/r/fent/node-ytdl-core/issues/479)

`related_videos` is always empty.

I'm using the latest version of ytdl-core, 0.29.5.

At some point earlier today, YouTube changed something on their end that means ytdl-core fails to find the data. This also affects the [Invidious](https://github.com/omarroth/invidious) project in exactly the same way.

Related video data is still available in the HTML, so just changing where you look for it should be enough to fix this.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $40.00)

- [issuehunt issuehunt](https://issuehunt.io/u/issuehunt) ($40.00)
### Submitted pull Requests
- [#481 Fixed related videos returning empty array](https://issuehunt.io/r/fent/node-ytdl-core/pull/481)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/fent/node-ytdl-core/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
---
IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

ashikmeerankutty:
Fixes #479 
Before
Screenshot 2019-08-30 at 11 53 18 PM
After
Screenshot 2019-08-30 at 11 53 48 PM

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

This pull request has been submitted to:
- [#479: related_videos is always empty](https://issuehunt.io/repos/21333232/issues/479)
---

IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

### DIFF
--- a/lib/info-extras.js
+++ b/lib/info-extras.js
@@ -127,18 +127,18 @@ exports.getRelatedVideos = (body) => {
   let jsonStr = util.between(body, '\'RELATED_PLAYER_ARGS\': ', '\n');
   let watchNextJson;
   try {
-    jsonStr = jsonStr.substring(0, jsonStr.length - 1)
-    jsonStr = JSON.parse(jsonStr)
-    watchNextJson = JSON.parse(jsonStr.watch_next_response)
+    jsonStr = jsonStr.substring(0, jsonStr.length - 1);
+    jsonStr = JSON.parse(jsonStr);
+    watchNextJson = JSON.parse(jsonStr.watch_next_response);
   }
   catch (err) {
-    return []
+    return [];
   }
-  let secondaryResults = watchNextJson.contents.twoColumnWatchNextResults.secondaryResults.secondaryResults.results
+  let secondaryResults = watchNextJson.contents.twoColumnWatchNextResults.secondaryResults.secondaryResults.results;
   let recomendedVideos = [];
   for (let result of secondaryResults) {
     if (result.compactVideoRenderer) {
-      let details = result.compactVideoRenderer
+      let details = result.compactVideoRenderer;
       try {
         recomendedVideos.push({
           id: details.videoId,
@@ -149,10 +149,10 @@ exports.getRelatedVideos = (body) => {
           short_view_count_text: details.shortViewCountText.simpleText,
           view_count: details.viewCountText.simpleText.replace(' views', ''),
           length_seconds: details.lengthText.simpleText
-        })
+        });
       }
       catch (err) {}
     }
   }
-  return recomendedVideos
+  return recomendedVideos;
 }
--- a/lib/info-extras.js
+++ b/lib/info-extras.js
@@ -137,7 +137,7 @@ exports.getRelatedVideos = (body) => {
     return {
       id: util.between(e, 'id=', '&'),
       shortViewCount: util.between(e, 'short_view_count_text=', '+views')
-    }
+    };
   })
   let secondaryResults = watchNextJson.contents.twoColumnWatchNextResults.secondaryResults.secondaryResults.results;
   let recomendedVideos = [];
@@ -147,12 +147,12 @@ exports.getRelatedVideos = (body) => {
       try {
         let viewCount = details.viewCountText.simpleText;
         if(!viewCount)
-          viewCount = details.viewCountText.runs[0].text.replace(' watching','')
+          viewCount = details.viewCountText.runs[0].text.replace(' watching', '');
         let shortViewCount = details.shortViewCountText.simpleText;
         if(shortViewCount === 'Recommended for you')
         {
           if(shortViewCount = rvs_params.find((elem) => elem.id === details.videoId))
-            shortViewCount = shortViewCount.shortViewCount
+            shortViewCount = shortViewCount.shortViewCount;
         }
         recomendedVideos.push({
           id: details.videoId,
@@ -160,8 +160,8 @@ exports.getRelatedVideos = (body) => {
           author: details.shortBylineText.runs[0].text,
           ucid: details.shortBylineText.runs[0].navigationEndpoint.browseEndpoint.browseId,
           author_thumbnail: details.channelThumbnail.thumbnails[0].url,
-          short_view_count_text: shortViewCount.replace(' views',''),
-          view_count: (viewCount !== 'Recommended for you') ? viewCount.replace(' views','') : shortViewCount.replace(' views',''),
+          short_view_count_text: shortViewCount.replace(' views', ''),
+          view_count: (viewCount !== 'Recommended for you') ? viewCount.replace(' views', '') : shortViewCount.replace(' views', ''),
           length_seconds: details.lengthText.simpleText
         });
       }
--- a/lib/info-extras.js
+++ b/lib/info-extras.js
@@ -138,7 +138,7 @@ exports.getRelatedVideos = (body) => {
       id: util.between(e, 'id=', '&'),
       shortViewCount: util.between(e, 'short_view_count_text=', '+views')
     };
-  })
+  });
   let secondaryResults = watchNextJson.contents.twoColumnWatchNextResults.secondaryResults.secondaryResults.results;
   let recomendedVideos = [];
   for (let result of secondaryResults) {
--- a/lib/info-extras.js
+++ b/lib/info-extras.js
@@ -138,18 +138,18 @@ exports.getRelatedVideos = (body) => {
   let recomendedVideos = [];
   for (let result of secondaryResults) {
     if (result.compactVideoRenderer) {
-      let videoDetails = {};
       let details = result.compactVideoRenderer
       try {
-        videoDetails["id"] = details.videoId
-        videoDetails["title"] = details.title.simpleText
-        videoDetails["author"] = details.shortBylineText.runs[0].text
-        videoDetails["ucid"] = details.shortBylineText.runs[0].navigationEndpoint.browseEndpoint.browseId
-        videoDetails["author_thumbnail"] = details.channelThumbnail.thumbnails[0].url
-        videoDetails["short_view_count_text"] = details.shortViewCountText.simpleText
-        videoDetails["view_count"] = details.viewCountText.simpleText.replace(" views", "")
-        videoDetails["length_seconds"] = details.lengthText.simpleText
-        recomendedVideos.push(videoDetails)
+        recomendedVideos.push({
+          "id": details.videoId,
+          "title": details.title.simpleText,
+          "author": details.shortBylineText.runs[0].text,
+          "ucid": details.shortBylineText.runs[0].navigationEndpoint.browseEndpoint.browseId,
+          "author_thumbnail": details.channelThumbnail.thumbnails[0].url,
+          "short_view_count_text": details.shortViewCountText.simpleText,
+          "view_count": details.viewCountText.simpleText.replace(" views", ""),
+          "length_seconds": details.lengthText.simpleText
+        })
       }
       catch (err) {}
     }
--- a/lib/info-extras.js
+++ b/lib/info-extras.js
@@ -133,12 +133,7 @@ exports.getRelatedVideos = (body) => {
   catch (err) {
     return [];
   }
-  let rvsParams = jsonStr.rvs.split(',').map((e) => {
-    return {
-      id: util.between(e, 'id=', '&'),
-      shortViewCount: util.between(e, 'short_view_count_text=', '+views')
-    };
-  });
+  let rvsParams = jsonStr.rvs.split(',').map((e) => qs.parse(e));
   let secondaryResults = watchNextJson.contents.twoColumnWatchNextResults.secondaryResults.secondaryResults.results;
   let recomendedVideos = [];
   for (let result of secondaryResults) {
@@ -153,7 +148,7 @@ exports.getRelatedVideos = (body) => {
         if (!shortViewCount.match(/^\d/.test()) || !shortViewCount)
         {
           if (shortViewCount = rvsParams.find((elem) => elem.id === details.videoId)) {
-            shortViewCount = shortViewCount.shortViewCount;
+            shortViewCount = shortViewCount.short_view_count_text;
           }
         }
         recomendedVideos.push({
--- a/lib/info-extras.js
+++ b/lib/info-extras.js
@@ -135,7 +135,7 @@ exports.getRelatedVideos = (body) => {
   }
   let rvsParams = jsonStr.rvs.split(',').map((e) => qs.parse(e));
   let secondaryResults = watchNextJson.contents.twoColumnWatchNextResults.secondaryResults.secondaryResults.results;
-  let recomendedVideos = [];
+  let videos = [];
   for (let result of secondaryResults) {
     let details = result.compactVideoRenderer;
     if (details) {
@@ -146,13 +146,13 @@ exports.getRelatedVideos = (body) => {
         }
         let shortViewCount = details.shortViewCountText.simpleText;
         let shortViewCountTest = /^\d/.test(shortViewCount);
-        if (!shortViewCountTest){
+        if (!shortViewCountTest) {
           let shortViewCountFromRVS = rvsParams.find((elem) => elem.id === details.videoId)
           if (shortViewCountFromRVS) {
             shortViewCount = shortViewCountFromRVS;
           }
         }
-        recomendedVideos.push({
+        videos.push({
           id: details.videoId,
           title: details.title.simpleText,
           author: details.shortBylineText.runs[0].text,
@@ -166,5 +166,5 @@ exports.getRelatedVideos = (body) => {
       catch (err) {}
     }
   }
-  return recomendedVideos;
+  return videos;
 }
--- a/lib/info-extras.js
+++ b/lib/info-extras.js
@@ -140,7 +140,7 @@ exports.getRelatedVideos = (body) => {
         videoDetails["id"] = result.videoId
         videoDetails["title"] = result.title.simpleText
         videoDetails["author"] = result.shortBylineText.runs[0].text
-        videoDetails["ucid"] = result.shortBylineText.runs[0].navigationEndpoint
+        videoDetails["ucid"] = result.shortBylineText.runs[0].navigationEndpoint.browseEndpoint.browseId
         videoDetails["author_thumbnail"] = result.channelThumbnail.thumbnails[0].url
         videoDetails["short_view_count_text"] = result.shortViewCountText.simpleText
         videoDetails["view_count"] = result.viewCountText.simpleText.replace(" views","") || 0
--- a/lib/info-extras.js
+++ b/lib/info-extras.js
@@ -145,9 +145,8 @@ exports.getRelatedVideos = (body) => {
           viewCount = details.viewCountText.runs[0].text;
         }
         let shortViewCount = details.shortViewCountText.simpleText;
-        let shortViewCountTest = (/^\d/).test(shortViewCount);
-        if (!shortViewCountTest)
-        {
+        let shortViewCountTest = /^\d/.test(shortViewCount);
+        if (!shortViewCountTest){
           if (shortViewCount = rvsParams.find((elem) => elem.id === details.videoId)) {
             shortViewCount = shortViewCount.short_view_count_text;
           }
@@ -159,7 +158,7 @@ exports.getRelatedVideos = (body) => {
           ucid: details.shortBylineText.runs[0].navigationEndpoint.browseEndpoint.browseId,
           author_thumbnail: details.channelThumbnail.thumbnails[0].url,
           short_view_count_text: shortViewCount.split(' ')[0],
-          view_count: ((/^\d/.test(viewCount)) ? viewCount : shortViewCount).split(' ')[0],
+          view_count: (/^\d/.test(viewCount) ? viewCount : shortViewCount).split(' ')[0],
           length_seconds: details.lengthText.simpleText
         });
       }
--- a/lib/info-extras.js
+++ b/lib/info-extras.js
@@ -146,13 +146,15 @@ exports.getRelatedVideos = (body) => {
     if (details) {
       try {
         let viewCount = details.viewCountText.simpleText;
-        if (!viewCount)
+        if (!viewCount) {
           viewCount = details.viewCountText.runs[0].text.replace(' watching', '');
+        }
         let shortViewCount = details.shortViewCountText.simpleText;
         if (!shortViewCount.match(/^\d/.test()) || !shortViewCount)
         {
-          if (shortViewCount = rvsParams.find((elem) => elem.id === details.videoId))
+          if (shortViewCount = rvsParams.find((elem) => elem.id === details.videoId)) {
             shortViewCount = shortViewCount.shortViewCount;
+          }
         }
         recomendedVideos.push({
           id: details.videoId,
--- a/lib/info-extras.js
+++ b/lib/info-extras.js
@@ -124,11 +124,32 @@ exports.getPublished = (body) => {
  * @return {Array.<Object>}
  */
 exports.getRelatedVideos = (body) => {
-  let jsonStr = util.between(body, '\'RELATED_PLAYER_ARGS\': {"rvs":', '},');
+  let jsonStr = util.between(body, '\'RELATED_PLAYER_ARGS\': ','\n');
   try {
-    jsonStr = JSON.parse(jsonStr);
+    jsonStr = jsonStr.substring(0,jsonStr.length-1);
+    jsonStr = JSON.parse(jsonStr)
+    watch_next_json = JSON.parse(jsonStr["watch_next_response"])
+    secondary_results = watch_next_json.contents.twoColumnWatchNextResults.secondaryResults.secondaryResults.results
+    recomendedVideos = [];
+    for(i in secondary_results)
+    {
+      let videoDetails = {};
+      if(secondary_results[i].compactVideoRenderer !== undefined)
+      {
+        let result = secondary_results[i].compactVideoRenderer
+        videoDetails["id"] = result.videoId
+        videoDetails["title"] = result.title.simpleText
+        videoDetails["author"] = result.shortBylineText.runs[0].text
+        videoDetails["ucid"] = result.shortBylineText.runs[0].navigationEndpoint
+        videoDetails["author_thumbnail"] = result.channelThumbnail.thumbnails[0].url
+        videoDetails["short_view_count_text"] = result.shortViewCountText.simpleText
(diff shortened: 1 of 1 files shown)


NightYoshi370:
Can you include a link to the author as well? It'd be really useful

ashikmeerankutty:
> Can you include a link to the author as well? It'd be really useful

ucid contains the youtube channel id. using the id you can create the channel URL.

lib/info-extras.js
```diff
@@ -124,11 +124,32 @@ exports.getPublished = (body) => {
...
+      {
+        let result = secondary_results[i].compactVideoRenderer
+        videoDetails["id"] = result.videoId
+        videoDetails["title"] = result.title.simpleText
+        videoDetails["author"] = result.shortBylineText.runs[0].text
+        videoDetails["ucid"] = result.shortBylineText.runs[0].navigationEndpoint.browseEndpoint.browseId
+        videoDetails["author_thumbnail"] = result.channelThumbnail.thumbnails[0].url
+        videoDetails["short_view_count_text"] = result.shortViewCountText.simpleText
+        videoDetails["view_count"] = result.viewCountText.simpleText.replace(" views","") || 0
+        videoDetails["length_seconds"] = result.lengthText.simpleText || "0.00"
```
fent: coverage report is showing that these safety defaults (the variables after `||`), don't get used. safe to delete;

lib/info-extras.js
```diff
@@ -124,11 +124,32 @@ exports.getPublished = (body) => {
...
-  let jsonStr = util.between(body, '\'RELATED_PLAYER_ARGS\': {"rvs":', '},');
+  let jsonStr = util.between(body, '\'RELATED_PLAYER_ARGS\': ','\n');
   try {
-    jsonStr = JSON.parse(jsonStr);
+    jsonStr = jsonStr.substring(0,jsonStr.length-1);
+    jsonStr = JSON.parse(jsonStr)
+    watch_next_json = JSON.parse(jsonStr["watch_next_response"])
+    secondary_results = watch_next_json.contents.twoColumnWatchNextResults.secondaryResults.secondaryResults.results
+    recomendedVideos = [];
+    for(i in secondary_results)
```
fent: use `for(let result of ...`

lib/info-extras.js
```diff
@@ -124,11 +124,32 @@ exports.getPublished = (body) => {
...
+        videoDetails["view_count"] = result.viewCountText.simpleText.replace(" views","") || 0
+        videoDetails["length_seconds"] = result.lengthText.simpleText || "0.00"
+        recomendedVideos.push(videoDetails)
+      }
+    }
+    return recomendedVideos
   } catch (err) {
     return [];
   }
-  return jsonStr.split(',').map((link) => qs.parse(link));
```
fent: move everything that won't throw out of the `try` block, only keeping `JSON.parse` there. if the reason for everything else being in there is because some of those variables might later be undefined, then put another `try` block inside the for loop so that when going through each item, one being broken won't mean returning an empty list

lib/info-extras.js
```diff
@@ -124,11 +124,32 @@ exports.getPublished = (body) => {
...
+        let result = secondary_results[i].compactVideoRenderer
+        videoDetails["id"] = result.videoId
+        videoDetails["title"] = result.title.simpleText
+        videoDetails["author"] = result.shortBylineText.runs[0].text
+        videoDetails["ucid"] = result.shortBylineText.runs[0].navigationEndpoint.browseEndpoint.browseId
+        videoDetails["author_thumbnail"] = result.channelThumbnail.thumbnails[0].url
+        videoDetails["short_view_count_text"] = result.shortViewCountText.simpleText
+        videoDetails["view_count"] = result.viewCountText.simpleText.replace(" views","") || 0
+        videoDetails["length_seconds"] = result.lengthText.simpleText || "0.00"
+        recomendedVideos.push(videoDetails)
```
fent: would prefer that `videoDetails` gets defined inline in this call. since this is the only place it's used.

lib/info-extras.js
```diff
@@ -124,11 +124,32 @@ exports.getPublished = (body) => {
...
 exports.getRelatedVideos = (body) => {
-  let jsonStr = util.between(body, '\'RELATED_PLAYER_ARGS\': {"rvs":', '},');
+  let jsonStr = util.between(body, '\'RELATED_PLAYER_ARGS\': ','\n');
   try {
-    jsonStr = JSON.parse(jsonStr);
+    jsonStr = jsonStr.substring(0,jsonStr.length-1);
+    jsonStr = JSON.parse(jsonStr)
+    watch_next_json = JSON.parse(jsonStr["watch_next_response"])
+    secondary_results = watch_next_json.contents.twoColumnWatchNextResults.secondaryResults.secondaryResults.results
+    recomendedVideos = [];
```
fent: use `let` on these

lib/info-extras.js
```diff
@@ -1,7 +1,7 @@
-const qs       = require('querystring');
-const url      = require('url');
+const qs = require('querystring');
+const url = require('url');
 const Entities = require('html-entities').AllHtmlEntities;
-const util     = require('./util');
+const util = require('./util');
```
fent: looks like you ran a linter on the whole file? can you revert this part?

lib/info-extras.js
```diff
@@ -124,11 +124,37 @@ exports.getPublished = (body) => {
...
  */
 exports.getRelatedVideos = (body) => {
-  let jsonStr = util.between(body, '\'RELATED_PLAYER_ARGS\': {"rvs":', '},');
+  let jsonStr = util.between(body, '\'RELATED_PLAYER_ARGS\': ', '\n');
+  let watchNextJson;
   try {
-    jsonStr = JSON.parse(jsonStr);
-  } catch (err) {
-    return [];
+    jsonStr = jsonStr.substring(0, jsonStr.length - 1)
```
fent: not needed since the last character is a `\n`

lib/info-extras.js
```diff
@@ -124,11 +124,37 @@ exports.getPublished = (body) => {
...
+    jsonStr = JSON.parse(jsonStr)
+    watchNextJson = JSON.parse(jsonStr["watch_next_response"])
```
fent: use the dot syntax when you can for non-dynamic properties
```suggestion
    watchNextJson = JSON.parse(jsonStr.watch_next_response)
```

lib/info-extras.js
```diff
@@ -124,11 +124,37 @@ exports.getPublished = (body) => {
...
+        videoDetails["length_seconds"] = details.lengthText.simpleText
+        recomendedVideos.push(videoDetails)
+      }
+      catch (err) {
+
+      }
+    }
+  }
+  return recomendedVideos
+}
```
fent: there's no newline at end of file, which makes github and linters complain

lib/info-extras.js
```diff
@@ -124,11 +124,37 @@ exports.getPublished = (body) => {
...
+        videoDetails["ucid"] = details.shortBylineText.runs[0].navigationEndpoint.browseEndpoint.browseId
+        videoDetails["author_thumbnail"] = details.channelThumbnail.thumbnails[0].url
+        videoDetails["short_view_count_text"] = details.shortViewCountText.simpleText
+        videoDetails["view_count"] = details.viewCountText.simpleText.replace(" views", "")
+        videoDetails["length_seconds"] = details.lengthText.simpleText
+        recomendedVideos.push(videoDetails)
+      }
+      catch (err) {
+
+      }
```
fent: shorten this to `catch (err) {}` for brevity

lib/info-extras.js
```diff
@@ -124,11 +124,37 @@ exports.getPublished = (body) => {
...
   }
-  return jsonStr.split(',').map((link) => qs.parse(link));
-};
+  catch (err) {
+    return []
+  }
+  let secondaryResults = watchNextJson.contents.twoColumnWatchNextResults.secondaryResults.secondaryResults.results
+  let recomendedVideos = [];
+  for (let result of secondaryResults) {
+    if (result.compactVideoRenderer !== undefined) {
```
fent: just `(result.compactVideoRenderer)` should be enough
```suggestion
    if (result.compactVideoRenderer) {
```

lib/info-extras.js
fent: something like

```js
recommendedVideos.push({
  id: details.videoId,
  ...
});
```

lib/info-extras.js
ashikmeerankutty: There is a ' , ' at the end that causes JSON.parse to throw an exception 
image

lib/info-extras.js
```diff
@@ -124,11 +124,35 @@ exports.getPublished = (body) => {
...
+      let details = result.compactVideoRenderer;
+      try {
+        recomendedVideos.push({
+          id: details.videoId,
+          title: details.title.simpleText,
+          author: details.shortBylineText.runs[0].text,
+          ucid: details.shortBylineText.runs[0].navigationEndpoint.browseEndpoint.browseId,
+          author_thumbnail: details.channelThumbnail.thumbnails[0].url,
+          short_view_count_text: details.shortViewCountText.simpleText,
+          view_count: details.viewCountText.simpleText.replace(' views', ''),
```
omarroth: `shortViewCountText` can be any of:
```javascript
{"shortViewCountText": {"simpleText": "2.5M views"}}
{"shortViewCountText": {"simpleText": "Recommended for you"}} // Doesn't appear to be handled
{"shortViewCountText": {"runs": [{"text": "47"}, {"text": " watching"}]} // Doesn't appear to be handled
```

For "Recommended for you", it's still possible to provide a string in `X views` by retrieving the corresponding object from `rvs` as part of the `jsonStr` in `getRelatedVideos`.

ashikmeerankutty: Thank you for the review. Can you please check the latest commit

lib/info-extras.js
```diff
@@ -124,11 +124,49 @@ exports.getPublished = (body) => {
...
     return [];
   }
-  return jsonStr.split(',').map((link) => qs.parse(link));
-};
+  let rvsParams = jsonStr.rvs.split(',').map((e) => {
+    return {
+      id: util.between(e, 'id=', '&'),
+      shortViewCount: util.between(e, 'short_view_count_text=', '+views')
+    };
+  })
```
fent: use `qs.parse()` like before. also, missing a semicolon

actually, why not keep using these parameters, it looks like they contain everything except `ucid`, not sure what this param is either. @omarroth what do you think?

lib/info-extras.js
```diff
@@ -124,11 +124,49 @@ exports.getPublished = (body) => {
...
+  let recomendedVideos = [];
+  for (let result of secondaryResults) {
+    let details;
+    if (details = result.compactVideoRenderer) {
+      try {
+        let viewCount = details.viewCountText.simpleText;
+        if(!viewCount)
+          viewCount = details.viewCountText.runs[0].text.replace(' watching', '');
+        let shortViewCount = details.shortViewCountText.simpleText;
+        if(shortViewCount === 'Recommended for you')
```
fent: using `rvs` would also be more reliable than checking for strings like these, since these are language dependent

ashikmeerankutty: Thank you for your review. I will check that out.

ashikmeerankutty: I think the rvs_params contains only the recommended videos not related videos @fent

ashikmeerankutty: image

These data are obtained from the rvs_params.

lib/info-extras.js
ashikmeerankutty: ucid contains the channel id @fent

ashikmeerankutty: also, view count is not present.

lib/info-extras.js
fent: ah, so it shows up in some weird format sometimes? for me it shows this

Screen Shot 2019-09-02 at 11 04 10 AM

anyway, if we use this data instead of `rvsParams`, we can check if the string contains a number at the start, with like `/^\d/` and then we can assume it's not "Recommended for you" if it matches.

> I think the rvs_params contains only the recommended videos not related videos

aren't those the same?

ashikmeerankutty: > aren't those the same?
From my observation, only a part of the ones obtained from `compactVideoRenderer` is available in the `rvs_params` only the ones with `shortViewCount` as `Recommended For You`

ashikmeerankutty: > anyway, if we use this data instead of `rvsParams`, we can check if the string contains a number at the start, with like `/^\d/` and then we can assume it's not "Recommended for you" if it matches.

I will update that

lib/info-extras.js
ashikmeerankutty: > use `qs.parse()` like before. also, missing a semicolon
is it necessary to use qs.parse()

fent: there's already a library that parses querystrings, why try to parse out the values yourself?

lib/info-extras.js
```diff
@@ -124,11 +124,49 @@ exports.getPublished = (body) => {
...
+    return {
+      id: util.between(e, 'id=', '&'),
+      shortViewCount: util.between(e, 'short_view_count_text=', '+views')
+    };
+  });
+  let secondaryResults = watchNextJson.contents.twoColumnWatchNextResults.secondaryResults.secondaryResults.results;
+  let recomendedVideos = [];
+  for (let result of secondaryResults) {
+    let details;
+    if (details = result.compactVideoRenderer) {
```
fent: put this statement above after declaring `details`, then check for `details` here. eslint warns about this, because it's common to confuse it for `==`

lib/info-extras.js
```diff
@@ -124,11 +124,49 @@ exports.getPublished = (body) => {
...
+      try {
+        let viewCount = details.viewCountText.simpleText;
+        if(!viewCount)
```
fent: space after `if`, curly braces around the statement

lib/info-extras.js
```diff
@@ -124,11 +124,49 @@ exports.getPublished = (body) => {
...
+          viewCount = details.viewCountText.runs[0].text.replace(' watching', '');
```
fent: due to the possibility of this string being different for every language, how about instead split it by spaces and get the first part?
```suggestion
          viewCount = details.viewCountText.runs[0].text.split(' ')[0];
```

lib/info-extras.js
```diff
@@ -124,11 +124,49 @@ exports.getPublished = (body) => {
...
+        let shortViewCount = details.shortViewCountText.simpleText;
+        if(!shortViewCount.match(/^\d/) || !shortViewCount)
```
fent: space after `if`. use `/^\d/.test()` since we don't care about the match results, only if it matches or not

lib/info-extras.js
```diff
@@ -124,11 +124,49 @@ exports.getPublished = (body) => {
...
+            shortViewCount = shortViewCount.shortViewCount;
+        }
+        recomendedVideos.push({
+          id: details.videoId,
+          title: details.title.simpleText,
+          author: details.shortBylineText.runs[0].text,
+          ucid: details.shortBylineText.runs[0].navigationEndpoint.browseEndpoint.browseId,
+          author_thumbnail: details.channelThumbnail.thumbnails[0].url,
+          short_view_count_text: shortViewCount.replace(' views', ''),
+          view_count: (viewCount.match(/^\d/)) ? viewCount.replace(' views','') : shortViewCount.replace(' views',''),
```
fent: no need for parens around the if statement. do the same to `.split()` it instead of replacing " views"

lib/info-extras.js
ashikmeerankutty: 👍

lib/info-extras.js
ashikmeerankutty: 👍

lib/info-extras.js
```diff
@@ -124,11 +124,46 @@ exports.getPublished = (body) => {
...
+  for (let result of secondaryResults) {
+    let details = result.compactVideoRenderer;
+    if (details) {
+      try {
+        let viewCount = details.viewCountText.simpleText;
+        if (!viewCount) {
+          viewCount = details.viewCountText.runs[0].text.replace(' watching', '');
+        }
+        let shortViewCount = details.shortViewCountText.simpleText;
+        if (!shortViewCount.match(/^\d/.test()) || !shortViewCount)
```
fent: `/^\d/.test` needs to be called by itself, not inside `.match()`...

I also think the check for `!shortViewCount` should go before the regexp test, but with `.test`, it's not needed

lib/info-extras.js
fent: not resolved yet

lib/info-extras.js
```diff
@@ -124,11 +124,46 @@ exports.getPublished = (body) => {
...
+        {
+          if (shortViewCount = rvsParams.find((elem) => elem.id === details.videoId)) {
```
fent: do the same thing here with storing the result in a variable and then checking the variable in the `if` statement

lib/info-extras.js
```diff
@@ -124,11 +124,46 @@ exports.getPublished = (body) => {
...
+          }
+        }
+        recomendedVideos.push({
+          id: details.videoId,
+          title: details.title.simpleText,
+          author: details.shortBylineText.runs[0].text,
+          ucid: details.shortBylineText.runs[0].navigationEndpoint.browseEndpoint.browseId,
+          author_thumbnail: details.channelThumbnail.thumbnails[0].url,
+          short_view_count_text: shortViewCount.split(' ')[0],
+          view_count: viewCount.match(/^\d/.test()) ? viewCount.split(' ')[0] : shortViewCount.split(' ')[0],
```
fent: use `.test()` here too.

don't repeat `.split(' ')[0]`, put the object that the conditional operator switches out in parenthesis and the `.split(...` outside it to prevent repeated code

lib/info-extras.js
ashikmeerankutty: Thank you for the suggestion, resolved

lib/info-extras.js
ashikmeerankutty: Thank you for the review changed as suggested.

lib/info-extras.js
```diff
@@ -124,11 +124,47 @@ exports.getPublished = (body) => {
...
+  for (let result of secondaryResults) {
+    let details = result.compactVideoRenderer;
+    if (details) {
+      try {
+        let viewCount = details.viewCountText.simpleText;
+        if (!viewCount) {
+          viewCount = details.viewCountText.runs[0].text;
+        }
+        let shortViewCount = details.shortViewCountText.simpleText;
+        let shortViewCountTest = (/^\d/).test(shortViewCount);
```
fent: no need for parenthesis around regex

lib/info-extras.js
```diff
@@ -124,11 +124,47 @@ exports.getPublished = (body) => {
...
+        if (!shortViewCountTest)
+        {
```
fent: put curly braces on same line as if statement

lib/info-extras.js
fent: not resolved

lib/info-extras.js
```diff
@@ -124,11 +124,47 @@ exports.getPublished = (body) => {
...
+          }
+        }
+        recomendedVideos.push({
+          id: details.videoId,
+          title: details.title.simpleText,
+          author: details.shortBylineText.runs[0].text,
+          ucid: details.shortBylineText.runs[0].navigationEndpoint.browseEndpoint.browseId,
+          author_thumbnail: details.channelThumbnail.thumbnails[0].url,
+          short_view_count_text: shortViewCount.split(' ')[0],
+          view_count: ((/^\d/.test(viewCount)) ? viewCount : shortViewCount).split(' ')[0],
```
fent: no parens needed around `.test()` call

lib/info-extras.js
ashikmeerankutty: changed please take a look @fent

cloudrac3r:
Seems like YouTube's changed something yet again. You should check again that this works.

lib/info-extras.js
```diff
@@ -124,11 +124,47 @@ exports.getPublished = (body) => {
...
+    let details = result.compactVideoRenderer;
+    if (details) {
+      try {
+        let viewCount = details.viewCountText.simpleText;
+        if (!viewCount) {
+          viewCount = details.viewCountText.runs[0].text;
+        }
+        let shortViewCount = details.shortViewCountText.simpleText;
+        let shortViewCountTest = /^\d/.test(shortViewCount);
+        if (!shortViewCountTest){
```
fent: space before `{`

lib/info-extras.js
```diff
@@ -124,11 +124,47 @@ exports.getPublished = (body) => {
...
+    watchNextJson = JSON.parse(jsonStr.watch_next_response);
+  }
+  catch (err) {
     return [];
   }
-  return jsonStr.split(',').map((link) => qs.parse(link));
-};
+  let rvsParams = jsonStr.rvs.split(',').map((e) => qs.parse(e));
+  let secondaryResults = watchNextJson.contents.twoColumnWatchNextResults.secondaryResults.secondaryResults.results;
+  let recomendedVideos = [];
```
fent: let's just name this `videos` so that there isn't an issue with calling these either related or recommended videos

fent:
> Seems like YouTube's changed something yet again. You should check again that this works.

I just checked with a few videos, seems to work

fent:
just noticed a few issues

Screen Shot 2019-09-07 at 11 05 36 PM

- `view_count` contains commas, that can easily be fixed with a `.replace()`.
- `length_seconds` should be in seconds, not in time format. you can parse this with the `humanStr()` function from the m3u8stream module. it's used here: https://github.com/fent/node-ytdl-core/blob/9aa2f4b9db73a47a0846a6282752f4293302045a/lib/index.js#L96 , it returns milliseconds btw
- `short_view_count_text` can be parsed too with `parseTime.durationStr()`, should add another property to the object and call it `short_view_count`

ashikmeerankutty:
> * `short_view_count_text` can be parsed too with `parseTime.durationStr()`, should add another property to the object and call it `short_view_count`

when I tried parseTime.durationStr() on short_view_count_text i'm getting these results

image

lib/info-extras.js
```diff
@@ -124,11 +124,49 @@ exports.getPublished = (body) => {
...
+        let shortViewCount = details.shortViewCountText.simpleText;
+        let shortViewCountTest = /^\d/.test(shortViewCount);
+        if (!shortViewCountTest) {
+          let shortViewCountFromRVS = rvsParams.find((elem) => elem.id === details.videoId)
+          if (shortViewCountFromRVS) {
+            shortViewCount = shortViewCountFromRVS;
+          }
+        }
+        viewCount = (/^\d/.test(viewCount) ? viewCount : shortViewCount).split(' ')[0]; 
+        let lengthSeconds = parseTime.humanStr(details.lengthText.simpleText);
```
fent: not sure how this is working here but, `parseTime` should be `require()`d at the top of the file

lib/info-extras.js
```diff
@@ -124,11 +124,49 @@ exports.getPublished = (body) => {
...
+        videos.push({
+          id: details.videoId,
+          title: details.title.simpleText,
+          author: details.shortBylineText.runs[0].text,
+          ucid: details.shortBylineText.runs[0].navigationEndpoint.browseEndpoint.browseId,
+          author_thumbnail: details.channelThumbnail.thumbnails[0].url,
+          short_view_count_text: shortViewCount.split(' ')[0],
+          view_count: viewCount.replace(',',''),
```
fent: space after `,` between arguments
```suggestion
          view_count: viewCount.replace(',', ''),
```

lib/info-extras.js
```diff
@@ -124,11 +124,49 @@ exports.getPublished = (body) => {
...
+          length_seconds: Math.floor(lengthSeconds/1000)
```
fent: spaces between math operators
```suggestion
          length_seconds: Math.floor(lengthSeconds / 1000)
```

lib/info-extras.js
fent: would also prefer if this call to `humanStr()` was placed inline in the line below, since `lengthSeconds` is only used there

fent:
> when I tried parseTime.durationStr() on short_view_count_text i'm getting these results
> 
> image

ah, sorry, my mistake. `durationStr()` is for parsing time strings. I'm ok with keeping it this way, since it was previously in this format anyway.

lib/info-extras.js
ashikmeerankutty: > not sure how this is working here but, `parseTime` should be `require()`d at the top of the file
sorry it wasn't working I was trying with another file, added the import

ashikmeerankutty: > would also prefer if this call to `humanStr()` was placed inline in the line below, since `lengthSeconds` is only used there

👍

lib/info-extras.js
```diff
@@ -124,11 +126,49 @@ exports.getPublished = (body) => {
...
+    if (details) {
+      try {
+        let viewCount = details.viewCountText.simpleText;
+        if (!viewCount) {
+          viewCount = details.viewCountText.runs[0].text;
+        }
+        let shortViewCount = details.shortViewCountText.simpleText;
+        let shortViewCountTest = /^\d/.test(shortViewCount);
+        if (!shortViewCountTest) {
+          let shortViewCountFromRVS = rvsParams.find((elem) => elem.id === details.videoId)
```
fent: missing a semicolon

lib/info-extras.js
```diff
@@ -124,11 +126,49 @@ exports.getPublished = (body) => {
...
+        }
+        viewCount = (/^\d/.test(viewCount) ? viewCount : shortViewCount).split(' ')[0];
+        videos.push({
+          id: details.videoId,
+          title: details.title.simpleText,
+          author: details.shortBylineText.runs[0].text,
+          ucid: details.shortBylineText.runs[0].navigationEndpoint.browseEndpoint.browseId,
+          author_thumbnail: details.channelThumbnail.thumbnails[0].url,
+          short_view_count_text: shortViewCount.split(' ')[0],
+          short_view_count: parseTime.durationStr(shortViewCount.split(' ')[0]),
```
fent: no longer used since `durationStr()` is for time not counts

fent:
thank you! and thank you for the patience with my comments

ashikmeerankutty:
:

> thank you! and thank you for the patience with my comments

Thank you for your comments, that helped me a lot 👍

PR #482, opened 2019-09-03
fix: add Error callback when fails to parse 'videoDetails'

### LINKED ISSUE #477: Cannot read property 'title' of undefined
Since today i get this error on all servers:

```
[...]/node_modules/ytdl-core/lib/info.js:192
data:    index.js:14600 -       title: info.player_response.videoDetails.title,
data:    index.js:14600 -                                               ^
data:    index.js:14600 - TypeError: Cannot read property 'title' of undefined
data:    index.js:14600 -     at err ([...]/node_modules/ytdl-core/lib/info.js:192:47)
data:    index.js:14600 -     at PassThrough.stream.on ([...]/node_modules/miniget/lib/index.js:195:30)
data:    index.js:14600 -     at emitNone (events.js:91:20)
data:    index.js:14600 -     at PassThrough.emit (events.js:185:7)
data:    index.js:14600 -     at endReadableNT (_stream_readable.js:978:12)
data:    index.js:14600 -     at _combinedTickCallback (internal/process/next_tick.js:80:11)
data:    index.js:14600 -     at process._tickCallback (internal/process/next_tick.js:104:9)
```

Btw.: How can i catch these error, so that my service not completely restart everytime it runs into it? 
Try - catch not working

luiseok:
resolves #477

## As is
ytdl-core crashes when youtube video is private.
If youtube video is private, there is no `videoDetails` key in `player_response`.
Therefore, an error occurs and is returned when you attempt to read the `.title` or `.lengthSeconds` property.
https://github.com/fent/node-ytdl-core/blob/0c5927c540499bfc2a59b9a5364dbace47dce57e/lib/info.js#L184-L194

## To be
I added `try{} catch(e) {}` to prevent the return of unhandled error while parsing videoDetails.

### DIFF
--- a/lib/info.js
+++ b/lib/info.js
@@ -179,7 +179,7 @@ const gotConfig = (id, options, additional, config, fromEmbed, callback) => {
       info.fmt_list.map((format) => format.split('/')) : [];
 
     info.formats = parseFormats(info);
-      // Add additional properties to info.
+    // Add additional properties to info.
     Object.assign(info, additional, {
       video_id: id,
 
@@ -187,7 +187,7 @@ const gotConfig = (id, options, additional, config, fromEmbed, callback) => {
       video_url: VIDEO_URL + id,
 
       // Copy over a few props from `player_response.videoDetails`
-      // for bakcwards compatibility.
+      // for backwards compatibility.
       title: info.player_response.videoDetails && info.player_response.videoDetails.title,
       length_seconds: info.player_response.videoDetails && info.player_response.videoDetails.lengthSeconds,
     });
--- a/lib/info.js
+++ b/lib/info.js
@@ -179,19 +179,23 @@ const gotConfig = (id, options, additional, config, fromEmbed, callback) => {
       info.fmt_list.map((format) => format.split('/')) : [];
 
     info.formats = parseFormats(info);
+    try {
+      // Add additional properties to info.
+      Object.assign(info, additional, {
+        video_id: id,
 
-    // Add additional properties to info.
-    Object.assign(info, additional, {
-      video_id: id,
+        // Give the standard link to the video.
+        video_url: VIDEO_URL + id,
 
-      // Give the standard link to the video.
-      video_url: VIDEO_URL + id,
+        // Copy over a few props from `player_response.videoDetails`
+        // for bakcwards compatibility.
+        title: info.player_response.videoDetails.title,
+        length_seconds: info.player_response.videoDetails.lengthSeconds,
+      });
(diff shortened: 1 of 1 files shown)


fent:
> ytdl-core crashes when youtube video is private.

this only happens when the video is private? is it still downloadable?

do you know if `videoDetails` is available anywhere else for such videos? I wouldn't wanna have these properties be undefined for private videos

luiseok:
@fent yes, it is still downloadable.
In any case, the videoDetails property cannot exist in the private video.

|Status|Downloadable|Fetchable(info)|
|:---:|:---:|:---:|
public|true|true
public(Disallow embedding)|true|true
not registered|true|true
region lock|false|true
Age-restricted|false|true
private|false|false
deleted|false|false
Youtube account terminated|false|false

fent:
ah that's interesting. age-restricted should be `true` for downloadable though.

lib/info.js
```diff
@@ -179,19 +179,23 @@ const gotConfig = (id, options, additional, config, fromEmbed, callback) => {
...
-      video_id: id,
+        // Give the standard link to the video.
+        video_url: VIDEO_URL + id,
 
-      // Give the standard link to the video.
-      video_url: VIDEO_URL + id,
+        // Copy over a few props from `player_response.videoDetails`
+        // for bakcwards compatibility.
+        title: info.player_response.videoDetails.title,
+        length_seconds: info.player_response.videoDetails.lengthSeconds,
```
fent: would rather check here if `videoDetails` is defined, rather than wrap this whole thing in a `try` block. otherwise, private videos will not get the other properties that do not depend on `videoDetails`, such as `video_id` and `video_url`

luiseok: Thanks! I totally agree.

luiseok: + the case which age-restricted couldn't downloadable was https://www.youtube.com/watch?v=bnb8-jpkQnc
It was `age-restricted` and `disallow embedding`.
sorry for mistake

lib/info.js
```diff
@@ -179,8 +179,7 @@ const gotConfig = (id, options, additional, config, fromEmbed, callback) => {
       info.fmt_list.map((format) => format.split('/')) : [];
 
     info.formats = parseFormats(info);
-
-    // Add additional properties to info.
+      // Add additional properties to info.
```
fent: this is still indented from the previous commit. please unindent

luiseok: got it, thanks

fent:
Thank you!

PR #500, opened 2019-10-01
test: fix test on related videos

TinyMan:
before we used `files/videos/3IqtmUscE_U-related/watch.html`
which contains no related videos.

also checks that `getRelatedVideos` returns a non empty array


test/files/refresh.js
```diff
@@ -150,7 +150,11 @@ const videos = [
     type: 'game',
     basicInfo: true,
     skip: ['get_video_info'],
-  }
+  },
+  {
+    id: '3IqtmUscE_U',
+    type: 'related',
+  },
```
fent: can you add `skip: ['get_video_info']` and maybe `'player'` in there too, since we mainly care about just the `watch.html` page

package.json
```diff
@@ -40,6 +40,7 @@
     "istanbul": "^0.4.5",
     "mocha": "^6.2.0",
     "muk-prop": "^2.0.0",
+    "muk-require": "^1.2.0",
```
fent: oops, must've forgotten to add this 😳

test/files/refresh.js
fent: also add `'player'` here so that the html5player file doesn't get saved

fent: oh wait it's not as simple as that, `skip` only matches exact filenames, and the player changes all the time. might change that later

TinyMan: Ok so .. I shouldn't add it right ? Since it wont work for now

fent: yeah it's fine for now. I'll add it in later

fent:
thanks again!

PR #504, opened 2019-10-12
Fixing rate limit and change formats source

RocoGreen:


RocoGreen:
I noticed a problem and i will fix it

RocoGreen:
Some errors occurrs and i need to retry with another pull request

PR #505, opened 2019-10-12
Change formats source

RocoGreen:
I know i already do a pull request but it was a problem with progress event and i fixed it.


RocoGreen:
This is the last edit but if there are a bug, i will fix it

lib/info.js
```diff
@@ -97,16 +97,18 @@ exports.getBasicInfo = (id, options, callback) => {
  * @return {Array.<Object>}
  */
 const parseFormats = (info) => {
-  let formats = [];
-  if (info.url_encoded_fmt_stream_map) {
-    formats = formats.concat(info.url_encoded_fmt_stream_map.split(','));
+  let formats = []
+
+  if (info.player_response.streamingData.formats) {
```
fent: is `streamingData` always defined? thinking that it should be checked if it's not

lib/info.js
```diff
@@ -97,16 +97,18 @@ exports.getBasicInfo = (id, options, callback) => {
...
+  formats = formats.concat(info.player_response.streamingData.formats)
```
fent: should be indented

lib/info.js
```diff
@@ -97,16 +97,18 @@ exports.getBasicInfo = (id, options, callback) => {
...
   }
-  if (info.adaptive_fmts) {
-    formats = formats.concat(info.adaptive_fmts.split(','));
+  if (info.player_response.streamingData.adaptiveFormats) {
+  formats = formats.concat(info.player_response.streamingData.adaptiveFormats)
```
fent: same comments here as above

lib/util.js
```diff
@@ -34,7 +34,7 @@ exports.sortFormats = (a, b) => {
 
   const getBitrate = (c) => {
     if (c.bitrate) {
-      let s = c.bitrate.split('-');
+      let s = c.bitrate.toString().split('-');
```
fent: is this a number now? if so, it wouldn't need to be `.split()`d anymore

RocoGreen: Sometimes, it's a number and sometimes not that's why i added `.toString()`

lib/info.js
RocoGreen: Ok, fixed

lib/info.js
RocoGreen: Fixed too

SwissCheese5:
Any updates on this?

fent:
merged in https://github.com/fent/node-ytdl-core/commit/eb1491f5f70302a2b11005041aa9cd0f4a53ef08

PR #526, opened 2019-11-24
add alert for recaptcha

mh4ck:
Nothing was thrown if the reCAPTCHA form is shown to the server and the process quit silently.

### DIFF
--- a/lib/info.js
+++ b/lib/info.js
@@ -123,7 +123,7 @@ exports.getBasicInfo = (id, options, callback) => {
 // };
 const parseFormats = (info, core_info) => {
   let formats = [];
-  if (core_info.csi_page_type === 'embed') { 
+  if (core_info.csi_page_type === 'embed' && core_info.player_response && core_info.player_response.streamingData) { 
       core_info.player_response.streamingData.formats.map(m => {
           let sp = '';
           let s = '';
--- a/lib/info.js
+++ b/lib/info.js
@@ -312,7 +312,9 @@ const getDashManifest = (url, options, callback) => {
   options.requestOptions = options.requestOptions || {};
   options.requestOptions.url = urllib.resolve(VIDEO_URL, url);
   const req = request(options.requestOptions);
-  req.setEncoding('utf8');
+  if(req.setEncoding) {
+    req.setEncoding('utf8');
+  }
   req.on('error', callback);
   req.on('data', (chunk) => { parser.write(chunk); });
   req.on('end', parser.close.bind(parser));
--- a/lib/info.js
+++ b/lib/info.js
@@ -108,64 +108,16 @@ exports.getBasicInfo = (id, options, callback) => {
  * @param {Object} info
  * @return {Array.<Object>}
  */
-// const parseFormats = (info) => {
-//   let formats = [];
-//   if (info.url_encoded_fmt_stream_map) {
-//     formats = formats.concat(info.url_encoded_fmt_stream_map.split(','));
-//   }
-//   if (info.adaptive_fmts) {
-//     formats = formats.concat(info.adaptive_fmts.split(','));
-//   }
-//   formats = formats.map((format) => querystring.parse(format));
-//   delete info.url_encoded_fmt_stream_map;
-//   delete info.adaptive_fmts;
-//   return formats;
-// };
-const parseFormats = (info, core_info) => {
+const parseFormats = (info) => {
   let formats = [];
-  if (core_info.csi_page_type === 'embed' && core_info.player_response && core_info.player_response.streamingData) { 
-      core_info.player_response.streamingData.formats.map(m => {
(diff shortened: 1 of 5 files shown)


TimeForANinja:
🤔 do i need to understand why you switched to request?

lib/info.js
```diff
@@ -54,6 +55,11 @@ exports.getBasicInfo = (id, options, callback) => {
       }
     }
 
+    const captchaForm = util.between(body, '<div class="g-recaptcha"', '>');
+    if(captchaForm) {
```
fent: if using a variable right away and only once, just put its value into where it's used

lib/info.js
```diff
@@ -54,6 +55,11 @@ exports.getBasicInfo = (id, options, callback) => {
...
+      return callback(Error("reCAPTCHA form shown"));
```
fent: use single quotes, and let's reword this to have a little more information on why the form was shown. is it shown when made too many requests, such as the 429 error? does the form say anything?

mh4ck:
> interested why you switched to request 🤔

Shit it shouldn't be in this Pull Request - also strange i'm pretty sure i made the pull request before i changed to request.

But to answer your question. We are not able to send proxy requests to YouTube with the current miniget implementation. With request it's working well.

lib/info.js
mh4ck: That's the HTML code of the form:

```
<div id="content" class="  content-alignment" role="main"><p class='largeText'>Sorry for the interruption. We have been receiving a large volume of requests from your network.</p>

<p>To continue with your YouTube experience, please fill out the form below.</p>
<form action="/das_captcha" method="POST">
<div class="g-recaptcha" data-sitekey="6Lf39AMTAAAAALPbLZdcrWDa8Ygmgk_fmGmrlRog"></div>
<br/>
<input type="hidden" name="next" value="/watch?v=s9xtPSweJNI&amp;hl=en&amp;bpctr=1574634970">
<input type="hidden" name="session_token" value="QUFFLUhqbE9talladHA2NmxZUk5CenQxZXU3WlQwSkYtUXxBQ3Jtc0ttNjB0UFgzRzU1bXRvU054VnJ2X01sMGh4Y3RKZHR0VUwtNldWZ045ZS1EbWFXbHJ1Rkh2TXYxTDJoNDljenFtOGtVbjlwZ3pTdktEZ3VkdjBFMnBBTEtGX0VqMUVUdzJ0NG13UV9ILUtzQzdvcW9IdTd6VFZTUzFfcjhJTWc2RF9BZUhXM3JSRVJENmJVMnBTSDVmaHBTUUJ6T1E=">
<input type="hidden" name="action_recaptcha_verify2" value="1">
<input type="submit" value="Submit">
</form>
```

lib/info.js
mh4ck: Isn't it more readable to use it like this. 
In my opinion there shouldn't be to much text in an if.

Also it's easy to debug by adding a simple console.log between this lines. 

I read it in a style guide but have forgotten which one, i think it's very usefull.

mh4ck:
> hi, thank you for taking the time to make the PR. besides the comments, I have two more things to address
> 
> * do you mind adding a test? look into the entries in `test/files/refresh.js`, although we can't easily replicate and get this error, we can add a `transform` in that file, adding in the error ourselves
> * I know you prefer request, but that's a separate change, and should be in it own PR if at all. we can have a discussion there about it

Sure i will try to write a test. :)
And also you are right, the change to request should be in another PR, what is the best practise to remove it from this PR?

fent:
> And also you are right, the change to request should be in another PR, what is the best practise to remove it from this PR?

`git rebase --interfactive` https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

then after you drop the commit(s), you can force-push your changes to this branch

lib/info.js
fent: how about the text from the `largeText` element?

lib/info.js
fent: hmm that's a good point

TimeForANinja: The problem with javascript is that since it's designed to be send over the web and interpreted live you (in general) want to reduce it as much as possible.
the performance should be about the same
removing the variable will end up with a line still easy to read and under 80 chars

lib/info.js
TimeForANinja: would be cool to export `data-sitekey` and `session_token` since (as far as i know) those two are everything that's needed to solve the captcha
maybe an optional `option.captchaSolver = (siteKey, sessionToken) => { ... }` ?

fent:
revert or remove the unwanted commits, instead of manually editing the files yourself and committing again.

use [`git rebase --interactive`](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History) to remove, or `git revert` to revert

PR #548, opened 2019-12-24
Fix Object.values error

### LINKED ISSUE #547: Type Error: info.js:240
I updated to latest version and got the following error a lot of times: 

```
/var/www/[project]/node_modules/ytdl-core/lib/info.js:240
data:    index.js:10046 -   info.formats = Object.values(formatsMap);
```

Fix: https://github.com/fent/node-ytdl-core/pull/548

After that i will create a PR for reCaptcha infos (already finished).

mh4ck:
Make it work on older node versions.

Fixes #547

### DIFF
--- a/lib/info.js
+++ b/lib/info.js
@@ -237,7 +237,7 @@ const mergeFormats = (info, formatsMap) => {
   info.formats.forEach((f) => {
     formatsMap[f.itag] = f;
   });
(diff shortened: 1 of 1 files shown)


fent:
are you on node v6? `Object.values` was added in node v8, which is the oldest node version the tests run on. I need to update the minimum node version supported in `package.json`

PR #572, opened 2020-02-15
Updated type definitions to add Captions and microformat fields

### LINKED ISSUE #571: Long links in videoInfo.descriptions are not usable and captions are also missing
Hi, I was using ytdl on javascript and didn't care about types, but recently I switched to typescript and found out that the only description provided in the type definitions is:
```videoInfo.description```, but the problem with that is that youtube makes long links abbreviated in it, so I looked in the ```getInfo``` response and found that there is another version of the description which has full links, it is under
```videoInfo.player_response.microformat.playerMicroformatRenderer.description.simpleText```.
Also ```videoInfo.player_response.captions``` are not included in the type definition.

I'm wondering, is there a reason not to include full response types in the type definitions? since it's returning the full data anyways.

Keep up the good work!

Aasim-A:
This PR is intended to close #571

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -225,7 +225,7 @@ declare module 'ytdl-core' {
           formats: {}[];
           adaptiveFormats: {}[];
         };
-        captions: {
+        captions?: {
           playerCaptionsTracklistRenderer: {
             captionTracks: captionTrack[];
           };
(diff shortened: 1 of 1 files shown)


Aasim-A:
I don't know how I missed it but I've found a mistake in the types, ```embed``` and all of the other properties should be inside ```playerMicroformatRenderer```, do I have to make another PR? or you can merge another commit from here

TimeForANinja:
gotta make another PR 😉

fent:
I missed that too ._.

PR #573, opened 2020-02-16
Fix ordering in microformat in type definitions

Aasim-A:
Recently I made a PR to add ```microformat``` to type definitions but there was a mistake that I didn't catch despite many reviews, so this PR fixes that by moving sub-fields of ```microformat``` under ```microformat.playerMicroformatRenderer```

sorry for that 😅


Aasim-A:
You're welcome!, you may release a new version quickly so that no one gets the previous build

PR #580, opened 2020-02-27
update ALL dependencies to latest

TimeForANinja:
update all dependencies
next step is to close all those Branches...


fent:
I need to publish muk-prop v2.0.0

other than that dependency, their latest still matches the semver version in `package.json`. the `^` will against against the latest from the given MAJOR version.

I can close those branches... yeah greenkeeper is noisy

PR #584, opened 2020-03-02
update hasAudio

iii8iii:
const hasAudio = format => !!Object.keys(format).filter(name=>{return /^audio/.test(name)}).length;


fent:
I think checking for `format.audioQuality` should work

PR #597, opened 2020-03-30
Add check for Captcha Form

mh4ck:
* Add check if captcha form exists on video page
* Throw an error with captcha form details

### DIFF
--- a/lib/info.js
+++ b/lib/info.js
@@ -35,7 +35,9 @@ exports.getBasicInfo = async (id, options) => {
     'User-Agent': '',
   });
 
+  console.log("[Info]: Start requesting info");
   let [, body] = await miniget.promise(url, reqOptions);
+  console.log("[Info]: Video info received");
 
   // Check if there are any errors with this video page.
   const unavailableMsg = util.between(body, '<div id="player-unavailable"', '>');
--- a/lib/info.js
+++ b/lib/info.js
@@ -49,7 +49,6 @@ exports.getBasicInfo = async (id, options) => {
   }
 
   // Check if there is any captcha form
-  console.log(body);
   const captchaForm = util.between(body, '<form action="/das_captcha" method="POST">', '</form>');
   if(captchaForm) {
     let details = {};
--- a/lib/info.js
+++ b/lib/info.js
@@ -35,9 +35,7 @@ exports.getBasicInfo = async (id, options) => {
     'User-Agent': '',
   });
 
-  console.log("[Info]: Start requesting info");
   let [, body] = await miniget.promise(url, reqOptions);
-  console.log("[Info]: Video info received");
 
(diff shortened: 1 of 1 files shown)


lib/info.js
```diff
@@ -48,6 +48,18 @@ exports.getBasicInfo = async (id, options) => {
     }
   }
 
+  // Check if there is any captcha form
+  const captchaForm = util.between(body, '<form action="/das_captcha" method="POST">', '</form>');
+  if(captchaForm) {
```
fent, line 53: lint: space before `{` please

lib/info.js
```diff
@@ -48,6 +48,18 @@ exports.getBasicInfo = async (id, options) => {
...
+    let details = {};
+    details.sitekey = util.between(captchaForm, 'data-sitekey="', '"');
+    details.next = util.between(captchaForm, 'name="next" value="', '"');
+    details.session_token = util.between(captchaForm, 'name="session_token" value="', '"');
+    details.action_recaptcha_verify2 = util.between(captchaForm, 'name="action_recaptcha_verify2" value="', '"');
+    details.largeText = util.between(body, '<p class=\'largeText\'">', '</p>');
+    throw Error("Captcha form received with info: " + JSON.stringify(details));
```
fent, line 60: is it better to print the json string of `details` here, or to attach all of these properties to the error object?

mh4ck:
> thanks for doing this. left a couple of comments.
> 
> I've been working on a change that uses a different endpoint than the watchpage, not sure how this is going to work with it. it might be that captcha is no longer an issue with the other endpoint since it gives a json response.

Oh that would be very nice, so you think we don't need this check any longer?
Then just remove this PR ^^

fent:
> Oh that would be very nice, so you think we don't need this check any longer?
> Then just remove this PR ^^

I don't know if it actually fixes it since I've never gotten the issue. the change is at
https://github.com/fent/node-ytdl-core/tree/new-endpoint

maybe you could try it and report if the captcha message is gone or displayed differently

TimeForANinja:
> > Oh that would be very nice, so you think we don't need this check any longer?
> > Then just remove this PR ^^
> 
> I don't know if it actually fixes it since I've never gotten the issue. the change is at
> https://github.com/fent/node-ytdl-core/tree/new-endpoint
> 
> maybe you could try it and report if the captcha message is gone or displayed differently

Did check it, downloaded all uploads (means ~50 requests to youtube to resolve playlists) from NoCopyrightSounds (744 videos each ~3.5MB audio only) twice and only had two statusCode `403`. Used my original 25 workers in parallel script - ending with ~40 Video downloads a minute. No more captcha for me...

fent:
superseded by https://github.com/fent/node-ytdl-core/pull/605

PR #599, opened 2020-04-06
better slice json, do not use javascript constants after it

zba:
This will fix problems when youtube code recompiled and after json in config something unexpected. 

fixes https://github.com/fent/node-ytdl-core/issues/585

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -356,56 +356,47 @@ exports.stripHTML = (html) => {
     .trim();
 };
 
-/** 
- * Match begin and end braces of input JSON, return only json 
+/**
+ * Match begin and end braces of input JSON, return only json
  * it works with braced jsons only!
  *
  * @param {String} mixedJson
  * @return {String}
 */
 exports.cutAfterJSON = function(mixedJson) {
-  let counter = 0,  isString, prevChar, open, close;
+  let open, close;
   if (mixedJson[0] === '[') {
     open = '[';
     close = ']';
   }
-  if (mixedJson[0] === '{') {
+  else if (mixedJson[0] === '{') {
     open = '{';
     close = '}';
   }
+
   if (!open) {
     throw new Error(`can't cut unsupported JSON (need to begin with [ or { ) but got ${mixedJson[0]}`);
   }
-  let result = '';
-  for (let character of mixedJson) {
-    result += character;
-    if (!isString && character === '"' && prevChar !== '\\') {
-      isString = true;
-      prevChar = character;
+
+  let isString = false, counter = 0;
+  for (let i = 0 ; i < mixedJson.length ; i++) {
+    if (mixedJson[i] === '"' && mixedJson[i-1] !== '\\') {
+      isString = !isString;
       continue;
     }
-    if (isString && character === '"' && prevChar !== '\\') {
-      isString = false;
-      prevChar = character;
-      continue
-    }
-    if (isString) {
-      prevChar = character;
-      continue;
-    }
-    if (prevChar === '\\') {
-      prevChar = character;
-      continue;
-    }
-    if (character === open) {
+
+    if (isString) continue;
+    if (mixedJson[i-1] === '\\') continue;
+
+    if (mixedJson[i] === open) {
       counter++;
     }
-    if (character === close) {
+    else if (mixedJson[i] === close) {
       counter--;
     }
+
     if (counter <= 0) {
-      return result;
+      return mixedJson.substr(0, i+1);
     }
-    prevChar = character;
   }
 }
--- a/lib/util.js
+++ b/lib/util.js
@@ -372,27 +372,37 @@ exports.cutAfterJSON = function(mixedJson) {
     close = '}';
   }
   if (!open) {
-    throw new Error("can't cut unsupported JSON (need to begin with [ or { )");
+    throw new Error("can't cut unsupported JSON (need to begin with [ or { ) but got " + mixedJson[0]);
   }
+  let result = '';
   for (let character of mixedJson) {
-    pos++;
+    result += character;
     if (!isString && character === '"' && prevChar !== '\\') {
       isString = true;
       prevChar = character;
       continue;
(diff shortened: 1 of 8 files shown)


lib/util.js
```diff
@@ -355,3 +355,55 @@ exports.stripHTML = (html) => {
     .replace(/<.*?>/gi, '')
     .trim();
 };
+
+/* Match begin and end braces of input JSON, return only json
```
TimeForANinja, line 359: doc-comments should start with a `/**`

lib/util.js
```diff
@@ -355,3 +355,55 @@ exports.stripHTML = (html) => {
...
+  if (mixedJson[0] === '[') {
+    open = '[';
+    close = ']';
+  }
+  if (mixedJson[0] === '{') {
+    open = '{';
+    close = '}';
+  }
+  if (!open) {
+    throw new Error("can't cut unsupported JSON (need to begin with [ or { ) but got " + mixedJson[0]);
```
TimeForANinja, line 375: try to stick with always using `'`

test/util-test.js
```diff
@@ -434,4 +434,33 @@ describe('util.addFormatMeta()', () => {
...
+      assert.equal(util.cutAfterJSON(
+        '{"a": "\\\"фыва", "b": 1, "c": {"test": 1}}abcd'),
+        '{"a": "\\\"фыва", "b": 1, "c": {"test": 1}}');
+    });
+    it('works with \\\\ in string', () => {
+      assert.equal(util.cutAfterJSON(
+        '{"a": "\\\\фыва", "b": 1, "c": {"test": 1}}abcd'),
+        '{"a": "\\\\фыва", "b": 1, "c": {"test": 1}}');
+    });
+  });
```
TimeForANinja, line 465: some newlines to match the rest would be appreciated

lib/util.js
```diff
@@ -355,3 +355,55 @@ exports.stripHTML = (html) => {
...
+      counter++;
+    }
+    if (character === close) {
+      counter--;
+    }
+    if (counter <= 0) {
+      return result;
+    }
+    prevChar = character;
+  }
```
TimeForANinja, line 408: first of all - i love the approach
just wanna throw sth in here:
i'd personally prefer an original `for (let i = 0 ; xxxx)` loop, removing the need for the `character` and `prevChar` variable (since we can do array-like access on strings) as well as the `result` variable which could be replaced with a single `.substring`

fent:
this is clever, I like 👍 

@TimeForANinja thanks for taking a look

lib/util.js
zba: I changed to accumulated string because there was some problem with slice result on some urls (cutted more, than needed)

test/util-test.js
zba: You mean make arguments on new lines like last 3 tests done ?

lib/util.js
zba: Would be nice to have eslint config, usually we not avoid double quotes if there is single quote inside.

lib/util.js
zba: prevChar is needed to correct parse entries like "\"" in strings, we need to ignore all braces inside string values...

zba:
I also think it would be better to make a function to extract jsons from js/html files and search ones which having expected content, it would fix many random errors  in parts, where json extracted.

lib/util.js
TimeForANinja: also love eslint
university is keeping me pretty busy thou...

test/util-test.js
TimeForANinja: i know whe're not 100% consistent but sth like a `\n` before the `describe` and some more between those `it`

lib/util.js
TimeForANinja: gave you an example in https://github.com/zba/node-ytdl-core/pull/1

lib/util.js
```diff
@@ -355,3 +355,48 @@ exports.stripHTML = (html) => {
...
+    }
+    else if (mixedJson[i] === close) {
+      counter--;
+    }
+
+    if (counter <= 0) {
+      return mixedJson.substr(0, i+1);
+    }
+  }
+}
```
fent, line 402: missing semicolon

lib/util.js
```diff
@@ -355,3 +355,55 @@ exports.stripHTML = (html) => {
     .replace(/<.*?>/gi, '')
     .trim();
 };
+
+/* Match begin and end braces of input JSON, return only json 
+ * it works with braced jsons only!
+ * @param {String} mixedJson
+ * @return {String}
+*/
+exports.cutAfterJSON = function(mixedJson) {
```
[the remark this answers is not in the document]
LevitatingBusinessMan: This should always return, but the return statement is conditioned?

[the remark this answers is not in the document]
fent: oh yeah, good catch. if the `counter` is over 0, then it's unclosed json, invalid, and should throw.

lib/util.js
```diff
@@ -355,3 +355,48 @@ exports.stripHTML = (html) => {
     .replace(/<.*?>/gi, '')
     .trim();
 };
+
+/**
+ * Match begin and end braces of input JSON, return only json
+ * it works with braced jsons only!
```
LevitatingBusinessMan, line 361: JSON is always braced.

lib/util.js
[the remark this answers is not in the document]
LevitatingBusinessMan: Alright am I being an idiot or does this function not actually cut the JSON at all, but is currently just verifying its integrity....`.lastIndexOf(';ytplayer.load')` is dropped, so this function is handling the uncut JSON...

I personally think it should cut it as soon as it faces a semi-colon with isString at false

lib/util.js
zba: quoted string and unquoted number, and boolean are valid JSON
`JSON.parse('1') // 1`

PR #607, opened 2020-04-12
change lint action

fent:
[hallee's action](https://github.com/hallee/eslint-action) kept getting a 401 error

    Error Error: Received status code 401

you can see it on this run
https://github.com/fent/node-ytdl-core/runs/581237974

their repo doesn't have issues opened, and the repo it was forked from hasn't been updated in over a year. so I searched for other eslint actions and found https://github.com/marketplace/actions/eslint-annotate

but because there's still the issue that it does not run on PRs from forks, I restored @TimeForANinja's commit
https://github.com/a-b-r-o-w-n/eslint-action/issues/6
https://github.com/Neovici/github-actions-eslint-annotator/issues/10

this is something that will likely be fixed for github actions eventually. but for now we can't use fancy annotated actions :(

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,6 +1,6 @@
 name: Lint
 
-on: pull_request
+on: [push, pull_request]
 
 jobs:
   eslint:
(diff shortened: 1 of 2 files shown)


TimeForANinja:
👍

PR #611, opened 2020-04-14
Improved version of #599

LevitatingBusinessMan:
I cleaned up some stuff. Should now merge and not throw ESLINT errors.

If @zba merges https://github.com/zba/node-ytdl-core/pull/2 than #599 should automatically update with my commit and be able to cleanly merge as well. If he doesn't than this can get merged instead.


LevitatingBusinessMan:
I just downloaded 527 video's with [yt_mp3](https://github.com/LevitatingBusinessMan/yt_mp3) and the only error I received was "Too many redirects" (once) compared to the 10+ JSON errors I normally receive.

test/util-test.js
```diff
@@ -434,4 +434,41 @@ describe('util.addFormatMeta()', () => {
...
+      assert.equal(util.cutAfterJSON('{"a": 1, "b": 1}'), '{"a": 1, "b": 1}');
+    });
+    it('Cut extra characters after JSON  end', () => {
+      assert.equal(util.cutAfterJSON('{"a": 1, "b": 1}abcd'), '{"a": 1, "b": 1}');
+    });
+    it('tolerant to string constants', () => {
+      assert.equal(util.cutAfterJSON('{"a": "}1", "b": 1}abcd'), '{"a": "}1", "b": 1}');
+    });
+    it('tolerant to string with escaped quoting', () => {
+      // eslint-disable-next-line no-useless-escape
```
fent, line 448: eslint is right with these warnings, the escapes aren't needed, the string will be the same

test/util-test.js
```diff
@@ -434,4 +434,41 @@ describe('util.addFormatMeta()', () => {
...
+      assert.equal(util.cutAfterJSON('{"a": "\\\"}1", "b": 1}abcd'), '{"a": "\\\"}1", "b": 1}');
+    });
+    it('works with nested', () => {
+      assert.equal(
+        // eslint-disable-next-line no-useless-escape
+        util.cutAfterJSON('{"a": "\\\"1", "b": 1, "c": {"test": 1}}abcd'),
+        // eslint-disable-next-line no-useless-escape, comma-dangle
```
fent, line 455: hmm, I don't know about disabling `comma-dangle`. I'll look into disabling this for function calls if it becomes too annoying

lib/util.js
```diff
@@ -374,3 +374,61 @@ exports.parseAbbreviatedNumber = string => {
...
+  if (mixedJson[0] === '[') {
+    open = '[';
+    close = ']';
+  } else if (mixedJson[0] === '{') {
+    open = '{';
+    close = '}';
+  }
+
+  if (!open) {
+    throw new Error(`can't cut unsupported JSON (need to begin with [ or { ) but got ${mixedJson[0]}`);
```
fent, line 396: use `throw Error` and capitalize the message (there isnt' an eslint rule for this)

also put a colon next to "got" for more clarity

lib/util.js
```diff
@@ -374,3 +374,61 @@ exports.parseAbbreviatedNumber = string => {
...
   return null;
 };
+
+/**
+ * Match begin and end braces of input JSON, return only json
+ *
+ * @param {String} mixedJson
+ * @return {String}
+*/
+// eslint-disable-next-line consistent-return
```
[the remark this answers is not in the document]
LevitatingBusinessMan: I think it is? The return value is in a conditional statement and eslint wouldn't stop screaming about it, even though it always returns. I'll give it a second look though.

LevitatingBusinessMan:
@fent
I was looking at the tests to add the ones you mentioned and I noticed that the way I wrote the function it shouldn't be able to handle some of the tests. The tests still passed though, or they seemed to have pasts even though they clearly didn't. Take a look at the actual logs: https://github.com/fent/node-ytdl-core/pull/611/checks?check_run_id=586823832#step:5:251
This means that the tests you got setup aren't functioning correctly.

test/util-test.js
```diff
@@ -434,4 +434,41 @@ describe('util.addFormatMeta()', () => {
...
+        '{"a": "\\\"фыва", "b": 1, "c": {"test": 1}}'
+      );
+    });
+    it('works with \\\\ in string', () => {
+      assert.equal(
+        util.cutAfterJSON('{"a": "\\\\фыва", "b": 1, "c": {"test": 1}}abcd'),
+        '{"a": "\\\\фыва", "b": 1, "c": {"test": 1}}',
+      );
+    });
```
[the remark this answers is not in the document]
LevitatingBusinessMan: I won't do one with uneven brackets but rather with missing closing brackets, due to the new way the function handles cutting the JSON now. The function should never encounter more closing brackets than opening brackets because the function starts with a single opening bracket and will cut the JSON as soon as it finds an equal amount of closing brackets.

One with a semi-colon ending would be unnecessary because there is already a "Cut extra characters after JSON end" test and the function now doesn't differentiate between a semi-colon and any other character anymore.

fent:
> @fent
> I was looking at the tests to add the ones you mentioned and I noticed that the way I wrote the function it shouldn't be able to handle some of the tests. The tests still passed though, or they seemed to have pasts even though they clearly didn't. Take a look at the actual logs: https://github.com/fent/node-ytdl-core/pull/611/checks?check_run_id=586823832#step:5:251
> This means that the tests you got setup aren't functioning correctly.

oh no, that's my fault. going to fix it now.

LevitatingBusinessMan:
When you change the CI tests, could you maybe also remove the codecov test? I have no idea what it does but it seems impossible to satisfy and because of it it's currently not possible to see if a PR passes directly from the PR page. Plus GitHub has been bombarding me with emails about the failed test.

Those CI tests from the last commit shouldn't appear when rerun.

test/util-test.js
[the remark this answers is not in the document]
fent: ok cool

lib/util.js
```diff
@@ -374,3 +374,63 @@ exports.parseAbbreviatedNumber = string => {
...
+      continue;
+    }
+    if (isString) continue;
+
+    if (mixedJson[i - 1] === '\\') continue;
+
+    // Cut at first semi-colon
+    if (mixedJson[i] === ';') {
+      break;
+    }
```
fent, line 419: can remove this since the function works differently now

lib/util.js
```diff
@@ -374,3 +374,63 @@ exports.parseAbbreviatedNumber = string => {
...
   return null;
 };
+
+/**
+ * Match begin and end braces of input JSON, return only json
+ *
+ * @param {String} mixedJson
+ * @return {String}
+*/
+
```
fent, line 384: remove extra line

lib/util.js
```diff
@@ -374,3 +374,63 @@ exports.parseAbbreviatedNumber = string => {
...
+  let i;
+  for (i = 0; i < mixedJson.length; i++) {
+    // Toggle the isString boolean when leaving/entering string
+    if (mixedJson[i] === '"' && mixedJson[i - 1] !== '\\') {
+      isString = !isString;
+      continue;
+    }
+    if (isString) continue;
+
+    if (mixedJson[i - 1] === '\\') continue;
```
fent, line 414: codecov shows this line is never reached. and that makes sense, there is only escaping inside strings, and if we're inside a string, we `continue` in the line above

LevitatingBusinessMan:
What's up with the codecov tests getting cancelled?
And that timeout should be increased or the test should be done async.

fent:
> What's up with the codecov tests getting cancelled?

they get cancelled when the tests fail. but that one failing test is not related to this PR.

PR #620, opened 2020-05-07
handle error

LeeHyKu:
When using ytdl(), all parameters are perfect, but sometimes errors occur, and this forces the program to shut down.

ps. I used a translator because of my poor English.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -105,6 +105,10 @@ const downloadFromInfoCallback = (stream, info, options) => {
         Range: `bytes=${options.range.start || '0'}-${options.range.end || ''}`,
       });
     }
+    if(format.url === undefined){
+      stream.emit('error', new Error('url is undefined'));
+      return;
+    }
 
     req = miniget(format.url, requestOptions);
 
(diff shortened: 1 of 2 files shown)


LeeHyKu:
@TimeForANinja yes

lib/info.js
```diff
@@ -303,7 +303,8 @@ for (let fnName of ['getBasicInfo', 'getFullInfo']) {
 
     if (callback) {
       return exports[fnName](link, options)
-        .then(info => callback(null, info), callback);
+        .then(info => callback(null, info), callback)
+        .catch(err => callback(err, null));
```
LeeHyKu, line 307: info.js:314:18/SyntaxError: Unexpected end of JSON input

lib/index.js
```diff
@@ -105,6 +105,10 @@ const downloadFromInfoCallback = (stream, info, options) => {
         Range: `bytes=${options.range.start || '0'}-${options.range.end || ''}`,
       });
     }
+    if (format.url === undefined) {
```
LeeHyKu, line 108: internal/validators.js:120:11/TypeError [ERR_INVALID_ARG_TYPE]: The "url" argument must be of type string. Received undefined

PR #634, opened 2020-05-25
remove duplicate formats from results

WaqasIbrahim:


WaqasIbrahim:
I noticed there were duplicate results for some videos that have dash manifest URL. But direct URL returns a 404 response and dash manifest formats were not replacing direct formats.

Examples:
https://www.youtube.com/watch?v=7W4o1Ca0bZo
https://www.youtube.com/watch?v=tEZ8uFPVSxE

WaqasIbrahim:
Not really as they are converted to strings when choosing a format
https://github.com/fent/node-ytdl-core/blob/5f0f802ed30343cb835e86cf8261fe63d9f095cb/lib/util.js#L147

But formats returned by YouTube have integer itags so the results should be uniform.

fent:
good point. alright then, thanks again!

PR #664, opened 2020-07-11
Fixes slow speed when downloading videoonly/audioonly formats

### LINKED ISSUE #294: Using audioonly filter slows down download tremendously
When downloading a video and converting to mp3 with ffmpeg, using `filter: 'audioonly'` in the options slows down speeds by magnitudes; ~72 seconds with filter, ~8 with no options, also ~8 with `quality: 'highestaudio'` (seems weird considering highestaudio requests just audio, like the filter).

This is tested using a [4:13 video](https://www.youtube.com/watch?v=sDLsSQf3Hc0) on my gigabit ethernet.

code used:
```js
const ffmpeg = require('fluent-ffmpeg');

let id = 'sDLsSQf3Hc0';
let start = Date.now();
let stream = ytdl(id, { quality: 'highestaudio'/*, filter: 'audioonly' */}); // mess around with quality, filter, and nothing at all being used

ffmpeg(stream)
	.on('end', () => {console.log('done, thanks - ' + (Date.now() - start) / 1000 + 's')})
	.audioBitrate(128)
	.save(`${__dirname}/${id}-${(Math.random() * 100000).toFixed(0)}.mp3`); // very efficient naming system here
```

This doesn't really affect me because I can just use `quality: 'highestaudio'`, but it does seem like a weird effect.

### LINKED ISSUE #327: 'audioonly' option not downloading properly
When I specify to only download the audio and pipe it into a .mp3 file, the file's download gets stuck super early on into the download and never continues downloading.

`ytdl(text, {
        quality: 'highestaudio',
        filter: 'audioonly'
      }).pipe(fs.createWriteStream(path + "/" + title + '.mp3'));`

where 'text' is the url.

Downloading as a video works perfectly and so does downloading the entire video w/o the audioonly and saving as an mp3 but doing that still includes the video data in the file.

### LINKED ISSUE #473: If `quality` param is set, ytdl(url, [options]) takes too long to download a video
I have experienced is that if it start downloading and **quality** param is set, it took too long to download the video comparing to the case when no **quality** param is specified.
I guess it's kinda some chunk stuff which causing this, but I need some clarifications.

I did an experiment to see the results and here's the outputs for each case.
### 1) When **quality** param is set to `lowest`:
```
> 0.83% downloaded - 0.02MB(s) of 1.89MB(s) - estimated time left: 1 second(s)
> 1.65 % downloaded - 0.03 MB(s) of 1.89 MB(s) - estimated time left: 2 second(s)
.....
> 98.46 % downloaded - 1.86 MB(s) of 1.89 MB(s) - estimated time left: 1 second(s)
> 99.29 % downloaded - 1.88 MB(s) of 1.89 MB(s) - estimated time left: 0 second(s)
> ✅ Video downloaded.It took: 105.319 second(s) for download process.
```

### 2) When **quality** param is NOT specified:
```
> 0.08% downloaded - 0.02MB(s) of 18.43MB(s) - estimated time left: 47 second(s)
> 0.17% downloaded - 0.03MB(s) of 18.43MB(s) - estimated time left: 40 second(s)
...
> 99.94% downloaded - 18.42MB(s) of 18.43MB(s) - estimated time left: 0 second(s)
✅ Video downloaded. It took: 18.737second(s) for download process.
```

As you can see, in case 1, it took almost 2 mins to download a 1.89MBs video, but in the case 2, it took just 18 seconds to download a 18.5MBs video.
As well as, in case 1, the progress event is fires for something like 5-8 times then it delays for ~1.5-2 seconds to fire again..

The code I used for this experiment:
```javascript
const fs = require("fs");
const ytdl = require("ytdl-core");
const path = require('path');

const output = path.resolve(__dirname, 'video-' + String(Date.now()) + '.mp4');

const video = ytdl("https://www.youtube.com/watch?v=4Tr0otuiQuU", {
  quality: "lowest"
});

let starttime = 0;
video.once("response", () => {
  starttime = Date.now();
});

video.on("progress", (chunkLength, downloaded, total) => {
  const percent = downloaded / total;
  const downloadedSeconds = (Date.now() - starttime) / 1000;

  const friedlyPercentage = Number((percent * 100).toFixed(2));
  const downloadedMbs = Number((downloaded / 1024 / 1024).toFixed(2));
  const totalInMbs = Number((total / 1024 / 1024).toFixed(2));
  const estimatedLeft = parseInt(downloadedSeconds / percent - downloadedSeconds, 10);

  const percentStr = `${friedlyPercentage}%`;
  const downloadedMbStr = `${downloadedMbs}MB(s) of ${totalInMbs}MB(s)`;
  const estimatedLeftStr = `${estimatedLeft} second(s)`;

  if (downloaded !== total)
  {
	console.log(
		`> ${percentStr} downloaded - ${downloadedMbStr} - estimated time left: ${estimatedLeftStr}`
	  );
  }
  else{
	  console.log(`✅ Video downloaded. It took: ${downloadedSeconds}second(s) for download process.`);
  }

});
```

### LINKED ISSUE #490: download progress slows down with filter of audioonly
I am converting youtube video to mp3 and then letting user download it but if

     ytdl(url,{
          filter: 'audioonly'
      })
      .pipe(res)
there is audioonly filter it downloading progress slows down. for example it video is length of 4:15, mp4 downloads size of 3x MB in five-six seconds and if it is audioonly size of 5MB, it takes 10-15sec or more. how can i fix this? here is full code:

    ytdl.getInfo(url, (err,info)=>{
           res.header('Content-Disposition', 'attachment; filename="'+ transliterate(info.title) 
     +'.mp3"');
           ytdl(url,{
              filter: 'audioonly'
           })
           .pipe(res)
    })

### LINKED ISSUE #660: Slow download for audio tracks
ytdl-core downloads the audio too slow, while others like ytdl-org/youtube-dl#21301 is just fine. Please look into this issue

Aasim-A:
When trying to download anything other than ```audio + video``` formats the download speed becomes extremely slow, that's because YouTube throttles the connection when trying to download at a high rate, so the solution to this is to download the video in chunks, I found the optimal chunk size to be ```10MB```, when trying to go above that the throttling starts.
Fixes #660 
Fixes #490 
Fixes #294 
Fixes #327 
Fixes #473

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ Attempts to download a video from the given url. Returns a [readable stream](htt
 * `liveBuffer` - How much time buffer to use for live videos in milliseconds. Default is `20000`.
 * `requestOptions` - Anything to merge into the request options which [miniget](https://github.com/fent/node-miniget) is called with, such as `headers`.
 * `highWaterMark` - How much of the video download to buffer into memory. See [node's docs](https://nodejs.org/api/stream.html#stream_constructor_new_stream_writable_options) for more. Defaults to 512KB.
-* `dlChunkSize` - The size of the download chunk in bytes, this is used when the chosen format is video only or audio only, the download in this case is separated into multiple chunks to avoid throttling. Defaults to 10MB.
+* `dlChunkSize` - The size of the download chunk in bytes. When the chosen format is video only or audio only, the download in this case is separated into multiple chunks to avoid throttling. Defaults to 10MB.
 * `lang` - The 2 character symbol of a language. Default is `en`.
 
 #### Event: info
--- a/README.md
+++ b/README.md
@@ -49,7 +49,6 @@ Attempts to download a video from the given url. Returns a [readable stream](htt
   ```
 * `format` - Primarily used to download specific video or audio streams. This can be a specific `format` object returned from `getInfo`.
   * Supplying this option will ignore the `filter` and `quality` options since the format is explicitly provided.
-* `dlChunkSize` - The size of the download chunk in bytes, this is used when the chosen format is video only or audio only, the download in this case is separated into multiple chunks to avoid throttling. Defaults to 10MB.
 * `range` - A byte range in the form `{start: INT, end: INT}` that specifies part of the file to download, ie {start: 10355705, end: 12452856}.
   * This downloads a portion of the file, and not a separately spliced video.
 * `begin` - What time in the video to begin. Supports formats `00:00:00.000`, `0ms, 0s, 0m, 0h`, or number of milliseconds. Example: `1:30`, `05:10.123`, `10m30s`.
@@ -58,6 +57,7 @@ Attempts to download a video from the given url. Returns a [readable stream](htt
 * `liveBuffer` - How much time buffer to use for live videos in milliseconds. Default is `20000`.
 * `requestOptions` - Anything to merge into the request options which [miniget](https://github.com/fent/node-miniget) is called with, such as `headers`.
 * `highWaterMark` - How much of the video download to buffer into memory. See [node's docs](https://nodejs.org/api/stream.html#stream_constructor_new_stream_writable_options) for more. Defaults to 512KB.
+* `dlChunkSize` - The size of the download chunk in bytes, this is used when the chosen format is video only or audio only, the download in this case is separated into multiple chunks to avoid throttling. Defaults to 10MB.
 * `lang` - The 2 character symbol of a language. Default is `en`.
 
 #### Event: info
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ Attempts to download a video from the given url. Returns a [readable stream](htt
   ```
 * `format` - Primarily used to download specific video or audio streams. This can be a specific `format` object returned from `getInfo`.
   * Supplying this option will ignore the `filter` and `quality` options since the format is explicitly provided.
+* `dlChunkSize` - The size of the download chunk in bytes, this is used when the chosen format is video only or audio only, the download in this case is separated into multiple chunks to avoid throttling. Defaults to `10485760` which is `10 Megabytes`.
 * `range` - A byte range in the form `{start: INT, end: INT}` that specifies part of the file to download, ie {start: 10355705, end: 12452856}.
   * This downloads a portion of the file, and not a separately spliced video.
 * `begin` - What time in the video to begin. Supports formats `00:00:00.000`, `0ms, 0s, 0m, 0h`, or number of milliseconds. Example: `1:30`, `05:10.123`, `10m30s`.
@@ -140,8 +141,6 @@ ytdl cannot download videos that fall into the following
 * Regionally restricted (requires a [proxy](example/proxy.js))
 * Private (if you have access, requires [cookies](example/cookie.js))
 * Rentals (if you have access, requires [cookies](example/cookie.js))
-
-YouTube intentionally rate limits downloads, particularly audio only formats, likely to prevent bandwidth abuse. The download rate is still faster than a media player can play the video, even on 2x. See [#294](https://github.com/fent/node-ytdl-core/issues/294).
 
 Generated download links are valid for 6 hours, for the same IP address.
 
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ Attempts to download a video from the given url. Returns a [readable stream](htt
   ```
 * `format` - Primarily used to download specific video or audio streams. This can be a specific `format` object returned from `getInfo`.
   * Supplying this option will ignore the `filter` and `quality` options since the format is explicitly provided.
-* `dlChunkSize` - The size of the download chunk in bytes, this is used when the chosen format is video only or audio only, the download in this case is separated into multiple chunks to avoid throttling. Defaults to `10485760` which is `10 Megabytes`.
+* `dlChunkSize` - The size of the download chunk in bytes, this is used when the chosen format is video only or audio only, the download in this case is separated into multiple chunks to avoid throttling. Defaults to 10MB.
 * `range` - A byte range in the form `{start: INT, end: INT}` that specifies part of the file to download, ie {start: 10355705, end: 12452856}.
   * This downloads a portion of the file, and not a separately spliced video.
 * `begin` - What time in the video to begin. Supports formats `00:00:00.000`, `0ms, 0s, 0m, 0h`, or number of milliseconds. Example: `1:30`, `05:10.123`, `10m30s`.
--- a/lib/index.js
+++ b/lib/index.js
@@ -178,7 +178,7 @@ const downloadFromInfoCallback = (stream, info, options) => {
     stream._isDestroyed = true;
     if (req.abort) req.abort();
     req.end();
-    req.removeAllListeners('data');
+    req.removeListener('data', ondata);
     req.unpipe();
   };
 };
--- a/lib/index.js
+++ b/lib/index.js
@@ -82,7 +82,7 @@ const downloadFromInfoCallback = (stream, info, options) => {
     stream.emit('progress', chunk.length, downloaded, contentLength);
   };
 
-  // Download the file in chunks, in this case the default 10MB,
+  // Download the file in chunks, in this case the default is 10MB,
   // anything over this will cause youtube to throttle the download
   const dlChunkSize = options.dlChunkSize || 1024 * 1024 * 10;
   let req;
@@ -114,7 +114,7 @@ const downloadFromInfoCallback = (stream, info, options) => {
     });
     pipeAndSetEvents();
   } else {
-    let requestOptions = Object.assign({}, options.requestOptions, {
+    const requestOptions = Object.assign({}, options.requestOptions, {
       maxReconnects: 6,
       maxRetries: 3,
       backoff: { inc: 500, max: 10000 },
@@ -127,9 +127,8 @@ const downloadFromInfoCallback = (stream, info, options) => {
       let end = start + dlChunkSize;
       const rangeEnd = options.range && options.range.end;
 
-      if (options.range) {
-        contentLength = (rangeEnd ? rangeEnd + 1 : format.contentLength) - start;
-      } else { contentLength = format.contentLength; }
+      contentLength = options.range ?
+        (rangeEnd ? rangeEnd + 1 : format.contentLength) - start : format.contentLength;
 
       const getNextChunk = () => {
         if (!rangeEnd && end >= contentLength) end = 0;
--- a/lib/index.js
+++ b/lib/index.js
@@ -87,8 +87,8 @@ const downloadFromInfoCallback = (stream, info, options) => {
 
   let req;
   let shouldEnd = true;
-  stream.on('initialize-stream', () => {
-    if (shouldEnd) stream.removeAllListeners('initialize-stream');
+  stream.on('set-pipe-and-events', () => {
+    if (shouldEnd) stream.removeAllListeners('set-pipe-and-events');
     // Forward events from the request to the stream.
     [
       'abort', 'request', 'response', 'error', 'retry', 'reconnect',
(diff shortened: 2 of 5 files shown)


fent:
hi, thanks for submitting the PR. I've inspected youtube's watch page when watching a video, and it indeed makes several ranged requests. I never looked into it enough to figure out a pattern, is 10mb the range used in these requests? was your process in figuring it out similar?

do you mind looking at [this PR on miniget](https://github.com/fent/node-miniget/pull/26). it implements the same functionality, but it's on the underlying http wrapper that ytdl-core uses. I'm not fully convinced that a "chunked" request feature belongs in miniget, but I'm more inclined towards that route. for ytdl-core, only the video request would use the feature, but I can see it used in other non-ytdl related requests. what do you think?

>downloading Dash format

dash mpd formats are downloaded using [m3u8stream](https://github.com/fent/node-m3u8stream/), which already uses segmented requests btw. I don't think you mean these though, there are other formats that are not in dashmpd format that are still labeled "dash" by youtube.

Aasim-A:
Hi,
> was your process in figuring it out similar?
I landed on the 10MB size after I tried a few speeds, you can set it to 11.5MB but I didn't think that was necessary and it might get throttled since it's on the edge of the allowed download rate, 10MB is accepted by all of the formats ranging from the lowest to the highest, and it's sufficient for high resolution downloading, if we choose a lower rate it might affect high resolution download speeds (eg 8K 60fps).

> I'm not fully convinced that a "chunked" request feature belongs in miniget, but I'm more inclined towards that route
Hmmm, an http client with that functionality would be nice, but is this chunked downloading supported for other file download types? (other that video and audio), and would that be suitable for 'miniget' ? since its concept is to be a minimalist get only http client, what are your thoughts?

> there are other formats that are not in dashmpd format that are still labeled "dash" by youtube
Aren't the audio only and video only formats labeled as 'Dash'? I thought they were called that.

fent:
> Hmmm, an http client with that functionality would be nice, but is this chunked downloading supported for other file download types? (other that video and audio), and would that be suitable for 'miniget' ? since its concept is to be a minimalist get only http client, what are your thoughts?

I'm imagining that it could be useful for other media hosts that also use chunk requests. but for now, I can't think of any, so maybe this feature is unique to youtube. if that's the case, it would be fine to put it in ytdl-core, and if we ever find other places to use it, we can port it over to miniget.

Aasim-A:
Sounds cool! I'm going to make some optimization to the code

lib/index.js
```diff
@@ -111,35 +129,70 @@ const downloadFromInfoCallback = (stream, info, options) => {
...
-    req = miniget(format.url, requestOptions);
-
-    req.on('response', res => {
-      if (stream._isDestroyed) { return; }
-      if (!contentLength) {
-        contentLength = parseInt(res.headers['content-length'], 10);
-      }
-    });
-    req.on('data', ondata);
+    const shouldBeChunked = !format.audioBitrate || !format.qualityLabel;
```
fent, line 132: use `format.hasAudio` and `format.hasVideo`. should also check that we're not chunking `format.isHLS` and `format.isDashMPD` since those are already segmented

lib/index.js
```diff
@@ -111,35 +129,70 @@ const downloadFromInfoCallback = (stream, info, options) => {
...
+    let start = (options.range && options.range.start) || 0;
+    let end = start + dlChunkSize;
+
+    if (shouldBeChunked) {
+      const heads = miniget(format.url);
```
fent, line 137: most formats now contain a `format.contentLength`, we don't need to make an extra request. below, I did check for the headers, just in case, and since the request was already made. but if we don't have to, let's avoid it.

the formats which don't have this property are segmented formats

lib/index.js
```diff
@@ -83,6 +86,20 @@ const downloadFromInfoCallback = (stream, info, options) => {
   };
 
   let req;
+  let shouldEnd = true;
+  stream.on('set-pipe-and-events', () => {
```
fent, line 90: make this a function `const pipeAndSetEvents = ...` rather than an event. this event wouldn't be used by users of this library, so it's not necessary to expose it.

lib/index.js
```diff
@@ -111,35 +129,70 @@ const downloadFromInfoCallback = (stream, info, options) => {
...
+        heads.abort();
+
+        const rangeEnd = options.range && options.range.end;
+
+        if (options.range) {
+          if (rangeEnd) contentLength = rangeEnd - start + 1;
+          else contentLength = parseInt(res.headers['content-length'], 10) - start;
+        } else { contentLength = parseInt(res.headers['content-length'], 10); }
+
+        const getChunk = () => {
```
fent, line 151: rename to `getNextChunk` so it's a little more verbose with what it's doing

lib/index.js
```diff
@@ -111,35 +129,70 @@ const downloadFromInfoCallback = (stream, info, options) => {
       });
```
fent, line 129: should remove this range request header

lib/index.js
Aasim-A: shouldn't the first part of the `if` statement handle `format.isHLS` and `format.isDashMPD`?

lib/index.js
```diff
@@ -44,6 +44,9 @@ const createStream = options => {
   return stream;
 };
 
+// Download the file in chunks, in this case 10MB, anything
+// over this will cause youtube to throttle the download
+const dlChunkSize = 10485760;
```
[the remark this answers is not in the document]
Aasim-A: Will it be better to add the ability to change the `DL_CHUNK_SIZE` through the options?

lib/index.js
Aasim-A: didn't understand 😕

lib/index.js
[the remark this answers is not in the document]
fent: maybe it's worth it. for tests, having it configurable will allow you to test with a smaller file, it would be nice not to have to commit a file that's over 10mb to the repo just for a test.

it could also be nice for users. frequently, users of the library will tweak the options little by little and finding out what works best for them. this has happened most recently with `highWaterMark`. we should provide sensible defaults, but also configurabilty, specially for an ever-changing unofficial public api.

lib/index.js
fent: oooh, good point!

lib/index.js
fent: there's a `Range` header above that is added if user has given `options.range`, it's overwritten later in chunked requests. so my initial suggestion should've been to move it to the `else` statement below, to where requests that are not chunked are handled

test/download-test.js
```diff
@@ -259,6 +351,83 @@ describe('Download video', () => {
...
+          scope.done();
+          assert.ok(equal);
+          done();
+        });
+      });
+    });
+
+    it('Chunks video only and chunk size matches given size', done => {
+      const dlChunkSize = 1024 * 200;
+      const stream = ytdl.downloadFromInfo(testInfo, { filter: format => !format.hasAudio, dlChunkSize });
```
fent, line 400: good opportunity to use `filer: 'videoonly'`

test/download-test.js
```diff
@@ -259,6 +351,83 @@ describe('Download video', () => {
...
+        const reqChunkSize = req.options.headers.range.split('-')[1];
+        assert.equal(reqChunkSize, dlChunkSize);
+        stream.removeAllListeners('request');
+        done();
+      });
+    });
+
+    it('Chunks audio only and chunk size matches given size', done => {
+      const dlChunkSize = 1024 * 200;
+      const stream = ytdl.downloadFromInfo(testInfo, { filter: format => !format.hasVideo, dlChunkSize });
```
fent, line 417: same here with `filter: 'audioonly'`

test/download-test.js
```diff
@@ -259,6 +351,83 @@ describe('Download video', () => {
...
+
+      stream.on('info', (info, format) => {
+        nock.url(format.url)
+          .reply(206);
+      });
+
+      stream.on('request', req => {
+        const reqChunkSize = req.options.headers.range.split('-')[1];
+        assert.equal(reqChunkSize, dlChunkSize);
```
fent, line 426: a number can equal a string? o_O

test/info-extras-test.js
```diff
@@ -91,6 +91,14 @@ describe('extras.getMedia()', () => {
...
     });
   });
+
+  describe('With invalid input', () => {
+    it('Should return an empty object', () => {
+      const media = extras.getMedia({ invalidObject: '' });
+      assert.ok(media);
+      assert.deepEqual(media, {});
+    });
+  });
```
fent, line 101: thanks for the extra test 👍

lib/index.js
```diff
@@ -96,50 +112,79 @@ const downloadFromInfoCallback = (stream, info, options) => {
...
+
+      if (options.range) {
+        if (rangeEnd) contentLength = rangeEnd - start + 1;
+        else contentLength = format.contentLength - start;
+      } else { contentLength = format.contentLength; }
+
+      const getNextChunk = () => {
+        if (!rangeEnd && end >= contentLength) end = 0;
+        if (rangeEnd && end > rangeEnd) end = rangeEnd;
+        shouldEnd = !end || end === rangeEnd;
```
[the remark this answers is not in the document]
Aasim-A: I have tested this, if `rangeEnd` is not explicitly specified, `contentLength` would mess up when `range` is given

test/download-test.js
Aasim-A: `filer: 'videoonly'` returns `video + audio` format 😕

README.md
```diff
@@ -49,6 +49,7 @@ Attempts to download a video from the given url. Returns a [readable stream](htt
   ```
 * `format` - Primarily used to download specific video or audio streams. This can be a specific `format` object returned from `getInfo`.
   * Supplying this option will ignore the `filter` and `quality` options since the format is explicitly provided.
+* `dlChunkSize` - The size of the download chunk in bytes, this is used when the chosen format is video only or audio only, the download in this case is separated into multiple chunks to avoid throttling. Defaults to 10MB.
```
fent, line 52: hmm, i think this should be moved down after `highWaterMark`. most used options should go near the top. and `dlChunkSize` is not something users will be changing unless they experience more throttling issues.

lib/index.js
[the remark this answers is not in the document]
Aasim-A: Added one more test and fixed a couple more to account for this 😊

README.md
```diff
@@ -57,6 +57,7 @@ Attempts to download a video from the given url. Returns a [readable stream](htt
 * `liveBuffer` - How much time buffer to use for live videos in milliseconds. Default is `20000`.
 * `requestOptions` - Anything to merge into the request options which [miniget](https://github.com/fent/node-miniget) is called with, such as `headers`.
 * `highWaterMark` - How much of the video download to buffer into memory. See [node's docs](https://nodejs.org/api/stream.html#stream_constructor_new_stream_writable_options) for more. Defaults to 512KB.
+* `dlChunkSize` - The size of the download chunk in bytes, this is used when the chosen format is video only or audio only, the download in this case is separated into multiple chunks to avoid throttling. Defaults to 10MB.
```
fent, line 60: can be shortened a bit
```suggestion
* `dlChunkSize` - The size of the download chunk in bytes. When the chosen format is video only or audio only, the download in this case is separated into multiple chunks to avoid throttling. Defaults to 10MB.
```

lib/index.js
```diff
@@ -96,50 +112,75 @@ const downloadFromInfoCallback = (stream, info, options) => {
...
+      pipeAndSetEvents();
+    }
   }
 
   stream.destroy = () => {
     stream._isDestroyed = true;
     if (req.abort) req.abort();
     req.end();
-    req.removeListener('data', ondata);
+    req.removeAllListeners('data');
```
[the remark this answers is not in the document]
Aasim-A: oh, I should've changed that, when i used a custom on data I did that but turns out I don't need to 🙂

PR #666, opened 2020-07-16
Working proxy example

Roki100:
So after testing the example with object (ip, port, auth) with around 10 proxies (5 with auth, 5 without auth) I've been always getting localhost errors such as `Error: connect ECONNREFUSED 127.0.0.1:8080` so Im updating the example to one that works


Roki100:
666 pr 😩

fent:
thank you!

PR #672, opened 2020-07-24
Update Discord example

promise:
• `message.member.voiceChannel` in discord.js v11 has been changed to `message.member.voice.channel` in v12.

• `dispatcher` no longer has an 'end'-event according to discord.js' documentation. But a 'speaking'-event was introduced in v12, this event gets triggered every time the speaking state changes. The variable `speaking` on line 26 is true if the bot is speaking, and false if it is no longer speaking. If the bot stops speaking, it leaves the voice channel.


fent:
thanks!

PR #674, opened 2020-07-28
Change assert.equal (deprecated) to assert.strictEqual

Aasim-A:
Make tests more strict 💪 .
`assert.equal` uses the `==` expression to evaluate (hence it's been deprecated), whereas `assert.strictEqual` uses `===` expression to evaluate.


fent:
thanks!

PR #676, opened 2020-07-29
add linter for typings & update integration

TimeForANinja:
[config is from the same place as the eslint config](https://github.com/discordjs/discord.js/blob/master/tslint.json)

no idea whether you're interested but was a fast pr 🤷

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "video",
     "download"
   ],
-  "version": "3.1.3",
+  "version": "3.2.0",
   "repository": {
     "type": "git",
     "url": "git://github.com/fent/node-ytdl-core.git"
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "ytdl-core",
-  "version": "3.1.3",
+  "version": "3.2.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {


fent:
looks great. I'm planning on converting the entire project to typescript, will this still be used then since there won't be a typings file?

TimeForANinja:
looks like it can handle typescript in general
but not 100% sure, since i know close to nothing about typescript

fent:
well we can include this at least until typescript conversion is done. some linting and type checking is better than none.

thanks!

package.json
```diff
@@ -41,14 +43,17 @@
...
     "mocha": "^7.0.0",
     "muk-prop": "^2.0.0",
     "muk-require": "^1.2.0",
     "nock": "^12.0.0",
     "nyc": "^15.0.0",
     "sinon": "^9.0.0",
-    "stream-equal": "~1.1.0"
+    "stream-equal": "~1.1.0",
+    "tslint": "^6.1.2",
+    "typescript": "^3.9.7"
```
fent, line 56: isn't dtslint the only needed dependency here, since it installs typescript and tslint according to its page https://www.npmjs.com/package/dtslint

TimeForANinja: kinda yes but kinda no
* it does install tslint as a dependencie
* typescript is a peer- & devDependencies
  those do NOT get installed automatically - that feature got remove ~2 years ago

TimeForANinja:
blaming this build-fail 10/10 on the package-lock.json
still not a friend of those

package.json
```diff
@@ -30,7 +30,9 @@
     "test:unit": "mocha --ignore test/irl-test.js test/*-test.js --timeout 4000",
     "test:irl": "mocha --timeout 16000 test/irl-test.js",
     "lint": "eslint ./",
-    "lint:fix": "eslint --fix ./"
+    "lint:fix": "eslint --fix ./",
+    "lint:typings": "tslint typings/index.d.ts",
+    "lint:typings:fix": "tslint --fix typings/index.d.ts"
```
fent, line 35: should now be updating to run `dtslint`, otherwise it fails

package.json
```diff
@@ -41,14 +43,16 @@
...
     "eslint": "^6.8.0",
     "mocha": "^7.0.0",
     "muk-prop": "^2.0.0",
     "muk-require": "^1.2.0",
     "nock": "^12.0.0",
     "nyc": "^15.0.0",
     "sinon": "^9.0.0",
-    "stream-equal": "~1.1.0"
+    "stream-equal": "~1.1.0",
+    "typescript": "^3.9.7"
```
fent, line 55: typescript can be removed too since it's also included in dtslint

TimeForANinja:
ok i don't understand that
my [test-branch](https://github.com/TimeForANinja/node-ytsr/tree/rework-testing) passes the build
this one with identical test & package.json doesn't

both dtslint and tslint do work
removing typescript still crashes
lemme real quick so that the integration also crashes 💯

TimeForANinja:
there we go
```Error: Cannot find module 'typescript'```

fent:
> there we go
> `Error: Cannot find module 'typescript'`

should be safe to put `typescript` back then

TimeForANinja:
ok, guess github didn't like that 😂

PR #678, opened 2020-07-31
Lint

TimeForANinja:
yet another try
went back to `tslint typings/typings.d.ts` since it produces way less & more understandable output
build should work
cleaner/new package-lock.json


fent:
thanks!

PR #713, opened 2020-09-28
feat: add IPv6 block rotating

### LINKED ISSUE #635: Preventing Error 429: Too Many Requests
I'm using ytdl-core from a browser, using browserify. Cross-domain requests are not allowed, so I'm using CORS Anywhere proxy to bypass CORS Header related errors.

After a few requests, I'm getting error 429s from YouTube, saying that I may be making too many requests (not really, one or two per minute at the most, downloading a webm stream). The method crashes afterward and won't do any additional requests.

Google seems to block the entire origin domain, so it won't work for anyone with the same origin header (I haven't verified yet).

There are multiple things wrong:
- How can I prevent this? I'd like to have some sort of rate throttling so it doesn't incur into 429 errors. It renders ytdl completely unusable for a long time (no exact amount, last time it was more than an hour)
- **No error** is thrown by ytdl whatsoever. This is not good because there is no way to handle the error. The callback never gets invoked, and error handlers are never called.

Code example:

```javascript
const concat = require('concat-stream');
 var concatStream = concat((arrayBuffer) =>
        {
            console.log('file downloaded');
           // Processing of video blob goes here...
        });

        var ytstream = ytdl(YOUTUBE_VIDEOURL,
            {
                requestOptions: {
                    transform: (parsed) =>
                    {  
                        var parsedHeaders = parsed.headers;
                        return {
                            host: 'cors-anywhere.herokuapp.com/', 
                            path: "http://youtube.com" + parsed.path,
                            maxRedirects: 10,
                            headers: parsedHeaders
                        }
                    },
                },
            });

        ytstream.on('error', function(e)
        {
           // Never gets called
        })
        .pipe(concatStream);
```

Roki100:
This feature was coded in a collab with @Million900o
Hello! I would like to present the new feature(s):
* Rotating trough IPv6 block to stop YouTube IP Bans (/58 block or larger recommended, however /64 should work aswell)
* Ability to disable chunking, as it is not needed while streaming in 1x speed (in for example Discord bots)

How does the method implemented work: in short, it uses different IPv6 address each time you download/request something via ytdl, for larger blocks(that include more that one /64) it also uses random /64 to make IP Bans impossible to get.

Feel free to propose changes, as this code is for sure not the cleanest

**Note:** The IPv6 rotating will work ***only*** if the block provided is owned by you and configured in the OS!
For setting up guide i recommend reading:
https://contabo.com/blog/adding-ipv6-connectivity-to-your-server
https://contabo.com/blog/configuring-additional-ip-addresses/
https://support.us.ovhcloud.com/hc/en-us/articles/360002296020-How-to-Configure-an-IPv6-Address-in-Ubuntu
Or googling for more tutorials.

***Also note:*** This functionality has been tested with /64, /58 and /48 blocks, and it works, however its a first version and i will keep improving that.
Adding new ytdl dependency was needed because its kinda complex/huge math to get a random ipv6 subnet.

closes #635 

###### Inspired by [Lavalink](https://github.com/Frederikam/Lavalink)'s rotator.


Roki100:
I would also give our test results about IP Bans (tested on a 6.5k+ servers discord music bot):
* **IPv4:** ~2 days
* **IPv6 /64:** none
* **IPv6 /58:** none
* **IPv6 /48:** none

TimeForANinja:
I guess you already spottet that the CI asks you to add tests for the new code lines 😉

Million900o:
How do we make tests?

Roki100:
Done

test/info-test.js
```diff
@@ -104,6 +104,19 @@ describe('ytdl.getInfo()', () => {
...
+          type: 'vevo',
+          player: true,
+        });
+        let info = await ytdl.getInfo(id, { IPv6Block: '2001:2::/48' });
+        await nock.url(info.formats[0].url).reply(function checkAddr() {
+          if (this.req.options.localAddress.includes(':')) scope.done();
+        });
+      });
+    });
+
```
TimeForANinja, line 119: "without ipv6" would be cool
i know that the default socket defaults to pad addresses to ipv6
e.g.:
`localhost` => `::1`
`127.0.0.1` => `::ffff:127.0.0.1`
`10.0.10.59` => `::ffff:10.0.10.59`

[net#isIPv6](https://nodejs.org/api/net.html#net_net_isipv6_input) should do the trick

Roki100: so like replace the `.includes` part to `net#isIPv6` right

TimeForANinja: 👍

Roki100: done, did a typo in commit `if -> of` but i guess its fine now

fent:
hi, thanks for the PR Roki and Million900o! it's pretty impressive that there are no bans with this on your bot during your testing. how long have you ran it for using the PR?

> * Rotating trough IPv6 block to stop YouTube IP Bans (/58 block or larger recommended, however /64 should work aswell)
> * Ability to disable chunking, as it is not needed while streaming in 1x speed (in for example Discord bots)

I'm going to start looking at this within the next day or so, but I will say that generally a PR is one change. keeping PRs small and contained to their scope not only makes the review process simpler, but it makes history cleaner, release changes easier to list, and if we ever have to `git bisect` or `git revert` we can do that per PR.

please separate the above two changes into 2 PRs. I'd keep this one, remove any changes about disabling chunking since those changes should be smaller than IPv6 rotating, and create another one with only the chunking changes. but also, chunking can be disabled with `dlChunkSize: Infinity`, or is that not the same?

Roki100:
> hi, thanks for the PR Roki and Million900o! it's pretty impressive that there are no bans with this on your bot during your testing. how long have you ran it for using the PR?
> 
> > * Rotating trough IPv6 block to stop YouTube IP Bans (/58 block or larger recommended, however /64 should work aswell)
> > * Ability to disable chunking, as it is not needed while streaming in 1x speed (in for example Discord bots)
> 
> I'm going to start looking at this within the next day or so, but I will say that generally a PR is one change. keeping PRs small and contained to their scope not only makes the review process simpler, but it makes history cleaner, release changes easier to list, and if we ever have to `git bisect` or `git revert` we can do that per PR.
> 
> please separate the above two changes into 2 PRs. I'd keep this one, remove any changes about disabling chunking since those changes should be smaller than IPv6 rotating, and create another one with only the chunking changes. but also, chunking can be disabled with `dlChunkSize: Infinity`, or is that not the same?

The bot currently runs on a /64 block for over a week now and there are no bans so far

Also, what do i do to remove chunking changes from this pr? like just remove them and make a commit or there is a command for that? I am not sure about the `dlChunkSize: Infinity` but `chunking: false` seems to be easier to understand and it completely "disables" the chunking part in the code

fent:
> Also, what do i do to remove chunking changes from this pr? like just remove them and make a commit or there is a command for that?

remove the changes for it and make a commit

>I am not sure about the dlChunkSize: Infinity but chunking: false seems to be easier to understand and it completely "disables" the chunking part in the code

how about `dlChunkSize: 0`?

Roki100:
> remove the changes for it and make a commit

will do in a sec
> how about `dlChunkSize: 0`?

Hmm wouldnt it make ytdl go crazy?
```js
      let start = (options.range && options.range.start) || 0;
      let end = start + dlChunkSize;
```
so like start and end will be the same

fent:
> Hmm wouldnt it make ytdl go crazy?
> 
> ```js
>       let start = (options.range && options.range.start) || 0;
>       let end = start + dlChunkSize;
> ```
> 
> so like start and end will be the same

oh i meant have it be turned off if it detects that `dlChunkSize === 0`

Roki100:
> oh i meant have it be turned off if it detects that `dlChunkSize === 0`

Hmmm, i might change it to that then i guess(with addition to readme), but first of all i have to remove chunking stuff from this pr and open second one

Roki100:
Okay, i think i've removed every change for chunking

lib/util.js
```diff
@@ -471,3 +472,22 @@ exports.deprecate = (obj, prop, value, oldPath, newPath) => {
...
+/**
+ * Gets random IPv6 Address from a block
+ *
+ * @param {string} block
+ * @returns {string}
+ */
+exports.getRandomIPv6 = block => {
+  let subnet = parseInt(block.split('/')[1]);
+  block = block.split('/')[0];
+  if (!subnet || subnet > 128 || subnet < 24) throw new Error('Invalid IPv6 subnet');
```
fent, line 486:
```suggestion
  if (!subnet || subnet > 128 || subnet < 24) throw Error('Invalid IPv6 subnet');
```

lib/util.js
```diff
@@ -471,3 +472,22 @@ exports.deprecate = (obj, prop, value, oldPath, newPath) => {
...
+  // Other errors will be thrown by ip6 itself
+  let ip;
+  if (subnet < 64) block = ip6.randomSubnet(block, subnet, 64, 1, true)[0];
+  if (subnet > 64) ip = ip6.randomSubnet(block, subnet, 128, 1, true)[0];
+  else ip = ip6.randomSubnet(block, 64, 128, 1, true)[0];
```
fent, line 491: can be combined into `let ip = ip6.randomSubnet(block, max(subnet, 64), ...);`

lib/util.js
```diff
@@ -471,3 +472,22 @@ exports.deprecate = (obj, prop, value, oldPath, newPath) => {
...
+ * @param {string} block
+ * @returns {string}
+ */
+exports.getRandomIPv6 = block => {
+  let subnet = parseInt(block.split('/')[1]);
+  block = block.split('/')[0];
+  if (!subnet || subnet > 128 || subnet < 24) throw new Error('Invalid IPv6 subnet');
+  // Other errors will be thrown by ip6 itself
+  let ip;
+  if (subnet < 64) block = ip6.randomSubnet(block, subnet, 64, 1, true)[0];
```
fent, line 489: not sure what this does

test/download-test.js
```diff
@@ -589,6 +590,17 @@ describe('Download video', () => {
     });
   });
 
+  describe('With IPv6 Block', () => {
+    it('Sends request with IPv6 address', done => {
+      const stream = ytdl.downloadFromInfo(testInfo, { IPv6Block: '2001:2::/48' });
+      stream.on('info', (info, format) => {
+        nock.url(format.url).reply(function checkAddr() {
+          if (net.isIPv6(this.req.options.localAddress)) done();
```
fent, line 598: should be an assertion `asesrt.ok(net.isIPv6(...))`

test/info-test.js
```diff
@@ -104,6 +105,19 @@ describe('ytdl.getInfo()', () => {
...
 
+    describe('With IPv6 Block', () => {
+      it('Sends request with IPv6 address', async() => {
+        const scope = nock(id, {
+          type: 'vevo',
+          player: true,
+        });
+        let info = await ytdl.getInfo(id, { IPv6Block: '2001:2::/48' });
+        await nock.url(info.formats[0].url).reply(function checkAddr() {
+          if (net.isIPv6(this.req.options.localAddress)) scope.done();
```
fent, line 116: `assert.ok()` here too. and place `scope.done()` after the `await`

lib/util.js
fent: not sure that this is covered in the tests either. wrap it in `{}` to see in it's caught. i'll make sure to add an eslint rule about this in the future.

fent (changes requested):

lib/util.js
TimeForANinja: yep
currently working on removing that ip6 dependencie and thought the same
picking a random address from a /48 subnet should result in the same as first picking a random /64 subnet and then a /128 addr

Roki100: not really, it should use different /64 each time, and /64 itself has trillion+ ips so getting ip from same /64 few times in a row would be too common on /48

TimeForANinja: thats not how stochastics work

first rolling /64 and then /128 blocks results in the same variance (in both /64 and /128) as directly rolling /128

it's not like directly rolling /128 gives you "worse randomness" for your ipv6 network part

Roki100: i mean random /128 from /48 gives random /128 from it
getting random /64 from /48 and then /128 from the /64 is the more efficent way to stop 429

TimeForANinja: i normally hate ppl who do sth like this but i really wanna proof my point 😅
so feel free to try for yourself
https://gist.github.com/TimeForANinja/8403d624306abad4dcafbed148ddbaf0
the standard deviation says it all

Roki100: So you are going to remove ip6 dependency? or do i update readme and its technically ready to merge?

TimeForANinja: the ip6 replacement is already kinda done

i'd finish & merge this pr first thou with ip6 (the way it is - maybe remove that /64 blocks first part ^^)
that way i still got some time to add tests & clean up my code
and i hate poking around in other peoples pr's

Roki100: Okay, i'll update readme after i will wake up then 👍

Roki100:
I think the pr is ready to be merged in current state 👀

README.md
```diff
@@ -142,11 +143,16 @@ ytdl cannot download videos that fall into the following
...
 Generated download links are valid for 6 hours, and may only be downloadable from the same IP address.
+Using IPv6 should not cause same IP address limitation.
 
 ## Handling Separate Streams
 
 Typically 1080p or better video does not have audio encoded with it. The audio must be downloaded separately and merged via an appropriate encoding library. `ffmpeg` is the most widely used tool, with many [Node.js modules available](https://www.npmjs.com/search?q=ffmpeg). Use the `format` objects returned from `ytdl.getInfo` to download specific streams to combine to fit your needs. Look at [example/ffmpeg.js](example/ffmpeg.js) for an example on doing this.
 
+## How does using an IPv6 block help?
+
+Using an IPv6 block is essentially using millions of different IPv6 addresses. Changing the external IP that you use is similar to proxying except without bypassing any restrictions such as a region lock. In a /64 IPv6 block, there are 18,446,744,073,709,551,616 single IPv6 addresses. This would allow you to have many requests without using the same IPv6 address more than once.
```
fent, line 154:
```suggestion
Using an IPv6 block is essentially having millions of IPv6 addresses at your request. Changing the external IP that you use is similar to using a proxy, except without bypassing restrictions such as a region lock. In a /64 IPv6 block, there are 18,446,744,073,709,551,616 unique IPv6 addresses. This would allow you to make each request with a different IPv6 address.
```

lib/util.js
fent: this is still not resolved right?

Roki100: Well you said you are not sure what this does

lib/util.js
```diff
@@ -479,3 +480,22 @@ exports.deprecate = (obj, prop, value, oldPath, newPath) => {
...
+exports.getRandomIPv6 = block => {
+  let subnet = parseInt(block.split('/')[1]);
+  block = block.split('/')[0];
+  if (!subnet || subnet > 128 || subnet < 24) throw Error('Invalid IPv6 subnet');
+  // Other errors will be thrown by ip6 itself
+  let ip;
+  if (subnet < 64) block = ip6.randomSubnet(block, subnet, 64, 1, true)[0];
+  if (subnet >= 64) ip = ip6.randomSubnet(block, subnet, 128, 1, true)[0];
+  else ip = ip6.randomSubnet(block, 64, 128, 1, true)[0];
+  return ip;
```
TimeForANinja, line 500:
```suggestion
  return ip6.randomSubnet(block, subnet, 128, 1, true)[0];
```

test/download-test.js
```diff
@@ -589,6 +590,18 @@ describe('Download video', () => {
...
+      const stream = ytdl.downloadFromInfo(testInfo, { IPv6Block: '2001:2::/48' });
+      stream.on('info', (info, format) => {
+        nock.url(format.url).reply(function checkAddr() {
+          assert.ok(net.isIPv6(this.req.options.localAddress));
+          done();
+        });
+      });
+    });
+  });
+
```
TimeForANinja, line 604:
```suggestion

  describe('Without IPv6 Block', () => {
    it('Sends request with (default) IPv4 address', done => {
      const stream = ytdl.downloadFromInfo(testInfo);
      stream.on('info', (info, format) => {
        nock.url(format.url).reply(function checkAddr() {
          assert.ok(net.isIPv4(this.req.options.localAddress));
          done();
        });
      });
    });
  });
```

TimeForANinja: disclaimer - did not try this locally

Roki100: ok, sorry for a bit of inactivity from my side but ill do testing for all those changes today/this week(to make sure everything works as expected), fix conflicts and then we good i guess

brennanwilkes:
Any information about when this PR will be implemented? Struggling with 429 errors.

Roki100:
Hello! Sorry for inactivity on this pr from my side, i know it should've been fixed months ago, but there we go, lets get it merged! 👀

lib/util.js
Roki100: changed in recent commits

test/download-test.js
Roki100: added in recent commit

lib/util.js
Roki100: I think this can be marked as resolved

lib/utils.js
```diff
@@ -168,3 +169,18 @@ exports.checkForUpdates = () => {
...
+ *
+ * @param {string} block
+ * @returns {string}
+ */
+exports.getRandomIPv6 = block => {
+  let subnet = parseInt(block.split('/')[1]);
+  block = block.split('/')[0];
+  if (!subnet || subnet > 128 || subnet < 24) throw Error('Invalid IPv6 subnet');
+  // Other errors will be thrown by ip6 itself
+  return ip6.randomSubnet(block, subnet, 128, 1, true)[0];
```
TimeForANinja, line 185: i see we're not first rolling the /64 net
like it

typings/index.d.ts
```diff
@@ -24,6 +24,8 @@ declare module 'ytdl-core' {
       begin?: string | number | Date;
       liveBuffer?: number;
       highWaterMark?: number;
+      lang?: string;
```
TimeForANinja, line 27: what's on with this?
didn't see any code related to that

test/full-info-test.js
```diff
@@ -39,6 +40,27 @@ describe('ytdl.getInfo()', () => {
...
+      });
+    });
+  });
+
+  describe('With invalid IPv6 Block', () => {
+    it('Should give an error', async() => {
+      const id = '_HSylqgVYQI';
+      await assert.rejects(ytdl.getInfo(id, { IPv6Block: '2001:2::/200' }), /Invalid IPv6 subnet/);
+    });
+  });
```
TimeForANinja, line 62: we could test the `getRandomIPv6` (but not here) function instead of doing the same `pass an ipv6` test (like above) twice

so that we do
a) test the whole usage via ytdl-core once
b) do white box tests only on the `getRandomIPv6` function for error like
  b.a) no subnet (`2001:2::`) or 
  b.b) subnet only (`/64`) or
  b.c) passing errors from `ip6.randomSubnet`

TimeForANinja (commented):
boy typing with a broken hand suxx
next time i'll send a voice memo

package.json
```diff
@@ -35,6 +37,8 @@
     "lint:typings:fix": "tslint --fix typings/index.d.ts"
   },
   "dependencies": {
+    "html-entities": "^1.3.1",
```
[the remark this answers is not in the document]
Roki100: oops, i think it got added by accident

lib/utils.js
Roki100: yea it was a requested change by you so i changed 👀

typings/index.d.ts
Roki100: im pretty sure that was an option in ytdl before so it persisted here

TimeForANinja:
if you wanna get rid of that ipv6 library 😉 
```js
const randomIP = (ip) => {
    const [addr, mask] = ip.split('/');
    // normalize inputs
    let base10Mask = parseInt(mask);
    const base10addr = normalizeIP(addr);
    // get random addr to pad with
    // using Math.random since we're not requiring high level of randomness
    const randomAddr = new Array(8).fill(1).map(x => Math.floor(Math.random() * 0xffff));
    // merge base10addr with randomAddr
    const mergedAddr = randomAddr.map((randomItem, idx) => {
        // calculate the amount of static bits
        const staticBits = Math.min(base10Mask, 16);
        // adjust the bitmask with the staticBits
        base10Mask -= staticBits;
        // calculate the bitmask
        // lsb makes the calculation way more complicated
        const mask = 0xffff - ((2**(16 - staticBits)) - 1);
        // combine base10addr and random
        return (base10addr[idx] & mask) + (randomItem & (mask ^ 0xffff));
    });
    // return new addr
    return mergedAddr.map(x => x.toString('16')).join(':');
}

const normalizeIP = (ip) => {
    // split by fill position
    const parts = ip.split('::').map(x => x.split(':'));
    // normalize start and end
    const partStart = parts[0] || [];
    const partEnd = parts[1] || [];
    // placeholder for full ip
    const fullIP = new Array(8).fill(0);
    // fill in start and end parts
    for (let i = 0; i < Math.min(partStart.length, 7); i++) {
        fullIP[i] = parseInt(partStart[i], 16) || 0;
    }
    for (let i = 0; i < Math.min(partEnd.length, 7) ; i++) {
        fullIP[7 - i] = parseInt(partEnd[i], 16) || 0;
    }
    return fullIP;
}
```

muhitrhn:
Can anyone help me to setup IPv6 on AWS console and setting it up with? (I already have all the codes in place in the latest version.)

I really get it working.

TimeForANinja:
@Roki100 can you have another look, pls

Roki100:
> @Roki100 can you have another look, pls

oh, yea sure, i have to come back from holiday tho as i dont have pc atm (3 days)

TimeForANinja:
sure - just tell me when you had a look 😉

lib/utils.js
```diff
@@ -181,3 +181,76 @@ exports.checkForUpdates = () => {
...
+// eslint-disable-next-line max-len
+const IPV6_REGEX = /^(([0-9a-f]{1,4}:)(:[0-9a-f]{1,4}){1,6}|([0-9a-f]{1,4}:){1,2}(:[0-9a-f]{1,4}){1,5}|([0-9a-f]{1,4}:){1,3}(:[0-9a-f]{1,4}){1,4}|([0-9a-f]{1,4}:){1,4}(:[0-9a-f]{1,4}){1,3}|([0-9a-f]{1,4}:){1,5}(:[0-9a-f]{1,4}){1,2}|([0-9a-f]{1,4}:){1,6}(:[0-9a-f]{1,4})|([0-9a-f]{1,4}:){1,7}(([0-9a-f]{1,4})|:))\/(1[0-1]\d|12[0-8]|\d{1,2})$/;
+/**
+ * Quick check for a valid IPv6
+ * The Regex only accepts a subset of all IPv6 Addresses
+ *
+ * @param {string} ip the IPv6 block in CIDR-Notation to test
+ * @returns {boolean} true if valid
+ */
+const isIPv6 = exports.isIPv6 = ip => Boolean(IPV6_REGEX.test(ip));
```
VoltrexMaster, line 230:
```suggestion
const isIPv6 = exports.isIPv6 = ip => IPV6_REGEX.test(ip);
```
`RegExp.prototype.test()` already returns a boolean.

Roki100:
@TimeForANinja aww, my bad on not responding, yeah everything seems to be fine 👍

Issue #716, opened 2020-10-02
Player Config not found every so often


fbagjug:
Hello, I'm on the latest version of YTDL-Core which is v3.3.0. I have saw other people open issues about this but my issue isn't exactly the same, my issue is that randomly it will say "Player config not found" but if I re-request it then it works fine. The fact it doesn't have a consistency makes it more annoying and harder to debug, just wondering if anyone researched this?

WeebHiroyuki:
I can reproduce this issue.

fent:
any videos this happens on?

WaqasIbrahim:
https://www.youtube.com/watch?v=8qu2TOR0b_o
https://www.youtube.com/watch?v=EUq-tNlCo7s
https://www.youtube.com/watch?v=IPX5U5iprTA
https://www.youtube.com/watch?v=K5yRqt8QFcI
https://www.youtube.com/watch?v=f3Lmx2Atkuo

All of these are music videos btw.

WeebHiroyuki:
For me, it just happened randomly.

fbagjug:
> any videos this happens on?

Sorry for the late reply, when I was debugging this for you I got limited by the 429 error so had to wait. But for me it seems random. It had happened a few times on the same song but just then I tried to play it again and it was fine for me so yeah it's just random.

PR #725, opened 2020-10-14
use regex to extract player config from embed page

WaqasIbrahim:
Fixes: https://github.com/fent/node-ytdl-core/issues/716

For some videos we get this response:
```
"PLAYER_CONFIG":{"attrs":{"height":"100%","width":"100%","id":"video-player"},
```
Instead of this
```
'PLAYER_CONFIG': {"attrs":{"id":"video-player","width":"100%","height":"100%"},
```
This happens randomly so there are no specific video examples for this.


fent:
thank you for looking into this and fixing it

PR #734, opened 2020-10-21
Extra parenthesis

Infernosniper:
There is an extra parenthesis on "stream.emit.bind(stream, 'error');" In the original code there are two closing parenthesis. This is causing many crashes in my code.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -16,7 +16,7 @@ const ytdl = (link, options) => {
   const stream = createStream(options);
   ytdl.getInfo(link, options).then(info => {
     downloadFromInfoCallback(stream, info, options);
-  }, stream.emit.bind(stream, 'error'));
+  }, stream.emit.bind(stream, 'error');
   return stream;
 };
 module.exports = ytdl;


Asaduji:
that extra parenthesis is from `.then()`, as you can see this isn't passing any check

fent:
how is your code crashing from it?

Infernosniper:
Yeah, never mind. 15 seconds after I sent that message I realized it wasn't extra. I've just been getting an error "unexpected token ) In JSON at position 0" and was hoping id easily find the fix. I'll send the full stack trace next time I get the error though.

Asaduji:
> Yeah, never mind. 15 seconds after I sent that message I realized it wasn't extra. I've just been getting an error "unexpected token ) In JSON at position 0" and was hoping id easily find the fix. I'll send the full stack trace next time I get the error though.

Are you on latest? That should be fixed

Infernosniper:
> > Yeah, never mind. 15 seconds after I sent that message I realized it wasn't extra. I've just been getting an error "unexpected token ) In JSON at position 0" and was hoping id easily find the fix. I'll send the full stack trace next time I get the error though.
> 
> Are you on latest? That should be fixed

I'm almost certain I tried installing the latest. I'll double check on the morning though.

PR #767, opened 2020-11-05
Updated Typings for relatedVideo

MatteZ02:
Seems like the response for relatedVideo has changed and did not match the old typings. Changed the typings to match the response i got. I tried multiple times and this was the only type of response i got.


fent:
thanks Matte!

PR #779, opened 2020-11-16
Alert if ytdl-core is old

redbrain:
From my experience in the YTDL support server, upgrading the package to latest seems to be something many users forget to do, and they encounter errors because of it.

If merged, this PR would add an alert in the console if ytdl-core is outdated, along with instructions to update.

This requires no extra dependencies and has been made as minimal as possible so as not to add overhead.
To function, it relies on the GitHub API.

To maintain this feature, all that is required is to update the tag ('v4.0.5') in the if statement with each new release.
This was written in 5 minutes, so please make edits if desired.

By the way, thank you for keeping this library up-to-date; it's quite a gift.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -8,6 +8,7 @@ const miniget = require('miniget');
 const m3u8stream = require('m3u8stream');
 const parseTime = require('m3u8stream/dist/parse-time');
 
+miniget("https://api.github.com/repos/fent/node-ytdl-core/releases/latest",{headers:{"User-Agent":"a/b"}}).text().then(a=>{"v4.0.5"!==JSON.parse(a).tag_name&&console.log("\x1B[31mWARNING:\x1B[0m ytdl-core is out of date! Update it using 'npm i ytdl-core'.")});
 
 /**
  * @param {string} link
--- a/lib/index.js
+++ b/lib/index.js
@@ -10,9 +10,11 @@ const parseTime = require('m3u8stream/dist/parse-time');
 
 const current = 'v4.0.5';
 const message = '\x1B[31mWARNING:\x1B[0m ytdl-core is out of date! Update it using "npm i ytdl-core".';
-miniget('https://api.github.com/repos/fent/node-ytdl-core/releases/latest', { headers: { 'User-Agent': 'a/b' } })
+miniget('https://api.github.com/repos/fent/node-ytdl-core/releases/latest', { headers: { 'User-Agent': 'a/b' } } )
   .text().then(response => {
-    if (JSON.parse(response).tag_name !== current) console.log(message);
+    if (JSON.parse(response).tag_name !== current) {
+      console.log(message);
+    }
 });
 
 /**
--- a/lib/index.js
+++ b/lib/index.js
@@ -10,12 +10,12 @@ const parseTime = require('m3u8stream/dist/parse-time');
 
 const current = 'v4.0.5';
 const message = '\x1B[31mWARNING:\x1B[0m ytdl-core is out of date! Update it using "npm i ytdl-core".';
-miniget('https://api.github.com/repos/fent/node-ytdl-core/releases/latest', { headers: { 'User-Agent': 'a/b' } } )
+miniget('https://api.github.com/repos/fent/node-ytdl-core/releases/latest', { headers: { 'User-Agent': 'a/b' } })
   .text().then(response => {
     if (JSON.parse(response).tag_name !== current) {
-      console.log(message);
+      console.error(message);
     }
-});
+  });
 
 /**
  * @param {string} link
--- a/lib/index.js
+++ b/lib/index.js
@@ -8,7 +8,7 @@ const miniget = require('miniget');
 const m3u8stream = require('m3u8stream');
 const parseTime = require('m3u8stream/dist/parse-time');
 
-const current = 'v4.0.5';
+const { version } = require('../package.json');
 const message = '\x1B[31mWARNING:\x1B[0m ytdl-core is out of date! Update it using "npm i ytdl-core".';
 miniget('https://api.github.com/repos/fent/node-ytdl-core/releases/latest', { headers: { 'User-Agent': 'a/b' } })
   .text().then(response => {
--- a/lib/index.js
+++ b/lib/index.js
@@ -8,7 +8,12 @@ const miniget = require('miniget');
 const m3u8stream = require('m3u8stream');
 const parseTime = require('m3u8stream/dist/parse-time');
 
-miniget("https://api.github.com/repos/fent/node-ytdl-core/releases/latest",{headers:{"User-Agent":"a/b"}}).text().then(a=>{"v4.0.5"!==JSON.parse(a).tag_name&&console.log("\x1B[31mWARNING:\x1B[0m ytdl-core is out of date! Update it using 'npm i ytdl-core'.")});
+const current = 'v4.0.5';
+const message = '\x1B[31mWARNING:\x1B[0m ytdl-core is out of date! Update it using "npm i ytdl-core".';
+miniget('https://api.github.com/repos/fent/node-ytdl-core/releases/latest', { headers: { 'User-Agent': 'a/b' } })
+  .text().then(response => {
+    if (JSON.parse(response).tag_name !== current) console.log(message);
+});
 
 /**
  * @param {string} link
--- a/lib/index.js
+++ b/lib/index.js
@@ -8,14 +8,6 @@ const miniget = require('miniget');
 const m3u8stream = require('m3u8stream');
 const parseTime = require('m3u8stream/dist/parse-time');
 
-const { version } = require('../package.json');
-const message = '\x1B[31mWARNING:\x1B[0m ytdl-core is out of date! Update it using "npm i ytdl-core".';
-miniget('https://api.github.com/repos/fent/node-ytdl-core/releases/latest', { headers: { 'User-Agent': 'a/b' } })
-  .text().then(response => {
-    if (JSON.parse(response).tag_name !== current) {
-      console.error(message);
-    }
-  });
 
 /**
  * @param {string} link
--- a/lib/info.js
+++ b/lib/info.js
@@ -451,9 +451,9 @@ miniget('https://api.github.com/repos/fent/node-ytdl-core/releases/latest', { he
       lastUpdateCheck = Date.now();
         if (JSON.parse(response).tag_name !== version) {
           console.warn('\x1b[33mWARNING:\x1B[0m ytdl-core is out of date! Update with "npm install ytdl-core@latest".');
-      }
-  }
-});
+        }
+    }
+  });
 
 // Export a few helpers.
 exports.validateID = urlUtils.validateID;
--- a/lib/info.js
+++ b/lib/info.js
@@ -426,6 +426,20 @@ const getM3U8 = async(url, options) => {
   return formats;
 };
 
+// Check for updates.
+const { version } = require('../package.json');
+lastUpdateCheck = Date.now();
+const checkForUpdates = () => {
+  if (!process.env.YTDL_NO_UPDATE && Date.now() - lastUpdateCheck >= 1000 * 60 * 60 * 12) {
+    miniget('https://api.github.com/repos/fent/node-ytdl-core/releases/latest', { headers: { 'User-Agent': 'a/b' } })
+      .text().then(response => {
+        if (JSON.parse(response).tag_name !== version) {
+          console.warn('\x1b[33mWARNING:\x1B[0m ytdl-core is out of date! Update with "npm install ytdl-core@latest".');
+        }
+      });
+  }
+}
+
 
 // Cache get info functions.
 // In case a user wants to get a video's info before downloading.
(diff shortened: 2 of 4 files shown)


Peppy8651:
Nice. If someone would merge this that would be great.

Peppy8651:
I'll test this out.

redbrain:
Looks like the linter hates me, but that's because code written to be small is naturally difficult to read.
Here's the code passed through the de-uglifier. If you'd like to use that instead, make the change.
```js
miniget("https://api.github.com/repos/fent/node-ytdl-core/releases/latest", {
    headers: {
        "User-Agent": "a/b"
    }
}).text().then(response => {
    if ("v4.0.5" !== JSON.parse(response).tag_name) {
        console.log("\x1B[31mWARNING:\x1B[0m ytdl-core is out of date! Update it using 'npm i ytdl-core'.");
    }
});
```

Peppy8651:
Yea I have Eslint too. I will make some changes so my linter doesn't give me an error or something.

Peppy8651:
Just finished changes that will probably fit with the linter.

Peppy8651:
These are the new changes I made. However, I get an error.
[ytdl-core.zip](https://github.com/fent/node-ytdl-core/files/5550483/ytdl-core.zip)
`Unhandled promise rejection: TypeError: Cannot read property 'player_response' of undefined
    at pipeline (C:\Users\Owner\Documents\BirdyBirdy\node_modules\ytdl-core\lib\info.js:126:19)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at async exports.getBasicInfo (C:\Users\Owner\Documents\BirdyBirdy\node_modules\ytdl-core\lib\info.js:39:14)
    at async Object.execute (C:\Users\Owner\Documents\BirdyBirdy\commands\play.js:41:14)`

redbrain:
After being unable to read the eslintrc for 2 minutes, I realised that lint will always fail since "no-console" is set.
fent'll have to give this the okay then, because console.logging is not okay with the computer.

take a look at the first commit for the original code, and the second one for more readable code.
3 and 4 were trying to figure out why the linter failed.

lib/index.js
```diff
@@ -8,6 +8,14 @@ const miniget = require('miniget');
 const m3u8stream = require('m3u8stream');
 const parseTime = require('m3u8stream/dist/parse-time');
 
+const current = 'v4.0.5';
+const message = '\x1B[31mWARNING:\x1B[0m ytdl-core is out of date! Update it using "npm i ytdl-core".';
+miniget('https://api.github.com/repos/fent/node-ytdl-core/releases/latest', { headers: { 'User-Agent': 'a/b' } })
```
fent, line 13: is the `User-Agent` header needed?

lib/index.js
```diff
@@ -8,6 +8,14 @@ const miniget = require('miniget');
 const m3u8stream = require('m3u8stream');
 const parseTime = require('m3u8stream/dist/parse-time');
 
+const current = 'v4.0.5';
```
fent, line 11: this can be imported from `package.json`
```suggestion
const { version } = require('../package.json');
```

lib/index.js
```diff
@@ -8,6 +8,14 @@ const miniget = require('miniget');
...
+const message = '\x1B[31mWARNING:\x1B[0m ytdl-core is out of date! Update it using "npm i ytdl-core".';
```
fent, line 12: how about yellow to signify more of a warning? red is used more with errors

and i would put `npm install ytdl-core@latest`. `install` to be more specific on what the command does. and `latest` because otherwise npm will install the version that satisfies `package.json`

fent (changes requested):
hi! this is a good idea, i've left some comments.

additionally
- could use some sort of rate limiting. maybe a local file that stores the last time the version was checked, and only check once a day at most. but i'd feel better about an in memory variable due to the point below
- could be moved into the `getInfo` functions, in case an app runs for a long time, they'd be able to check for a new version whenever they call `getInfo()`, rate limited of course.
- should have an option to disable this. i've seen a few occasions of people using an older version for "reasons". i woudn't recommend it, but you can't stop someone from using the program how they like. maybe `checkUpdates/Release` or `skip(Update/Release)Check`?

lib/index.js
fent: oh, move this to the line below since it's only used there

lib/index.js
redbrain: while I was testing this, I got 403s from the service if it wasn't there. weird, but any valid user-agent fixed it.
Using node-fetch seems to solve that entirely, but i didnt want to add another dependency.

lib/index.js
```diff
@@ -8,6 +8,14 @@ const miniget = require('miniget');
...
+miniget('https://api.github.com/repos/fent/node-ytdl-core/releases/latest', { headers: { 'User-Agent': 'a/b' } })
+  .text().then(response => {
+    if (JSON.parse(response).tag_name !== current) {
+      console.error(message);
```
fent, line 16: i'd use `console.warn` since this is more of a warning. in node, it's the same as `console.error`, but in a browser it's printed differently

you can add `warn` to eslint's `noconsole` option https://eslint.org/docs/rules/no-console#options

fent:
> * should have an option to disable this. i've seen a few occasions of people using an older version for "reasons". i woudn't recommend it, but you can't stop someone from using the program how they like. maybe `checkUpdates/Release` or `skip(Update/Release)Check`?

or maybe and environment variable to more easily disable for tests

lib/index.js
redbrain: this has been done in the most recent commit

lib/index.js
redbrain: this has been done in the most recent commits

redbrain (commented):
The things you have requested should now be done in patch-1. Let me know if there's more you'd like to see or something I've missed. Feel free to change anything obvious.

lib/info.js
```diff
@@ -330,7 +330,19 @@ const parseFormats = player_response => {
  * @param {Object} options
  * @returns {Promise<Object>}
  */
+const { version } = require('../package.json');
+let lastUpdateCheck = Date.now();
 exports.getInfo = async(id, options) => {
+  update = process.env.UPDATE;
```
fent, line 336: how about `process.env.YTDL_NO_UPDATE`? just `UPDATE` may be too common and used by other programs

lib/info.js
```diff
@@ -330,7 +330,19 @@ const parseFormats = player_response => {
...
+  miniget('https://api.github.com/repos/fent/node-ytdl-core/releases/latest', { headers: { 'User-Agent': 'a/b' } })
+    .text().then(response => {
+      if (update === 'always' || (update !== 'never' && Date.now() - lastUpdateCheck >= 43200000)) {
```
fent, line 339: include the `process.env` code here to make all this a bit simpler to grasp
```js
if (!process.env.YTDL_NO_UPDATE && ...
```

this may make tests fail since it'll make a request on the first test, and requests are blocked in tests. you can put the following in `test/nock.js` (which is imported by all tests) to prevent that
```js
process.env.YTDL_NO_UPDATE = true;
```

if you feel up for it, feel free to write a test for this. if not, i can do it after the PR is merged.

fent, line 339: would prefer if the number here was not precalculated so that someone looking at this can know how much time it is. `43200000` -> `1000 * 60 * 60 * 12`

fent (changes requested):

redbrain:
Would someone mind fixing the code formatting according to eslint?
I'm on a different device so I'm using GH web and I have no idea what the linter is demanding of me now.

Peppy8651:
Ok, here's some changes to the current one that I made so it fits with my linter. Don't know how Github will handle it though.
[ytdl-core.zip](https://github.com/fent/node-ytdl-core/files/5554410/ytdl-core.zip)

lib/info.js
```diff
@@ -443,6 +443,17 @@ for (let funcName of ['getBasicInfo', 'getInfo']) {
...
 
+const { version } = require('../package.json');
+let lastUpdateCheck = Date.now();
+miniget('https://api.github.com/repos/fent/node-ytdl-core/releases/latest', { headers: { 'User-Agent': 'a/b' } })
+  .text().then(response => {
+    if (!process.env.YTDL_NO_UPDATE && Date.now() - lastUpdateCheck >= 1000 * 60 * 60 * 12) {
+      lastUpdateCheck = Date.now();
+        if (JSON.parse(response).tag_name !== version) {
+          console.warn('\x1b[33mWARNING:\x1B[0m ytdl-core is out of date! Update with "npm install ytdl-core@latest".');
+        }
```
fent, line 454: i think the indentation errors are from this line, it's 8 but should be 6

not sure why the linter is giving the other lines as errors o_O

lib/info.js
```diff
@@ -443,6 +443,17 @@ for (let funcName of ['getBasicInfo', 'getInfo']) {
   };
 }
 
+const { version } = require('../package.json');
+let lastUpdateCheck = Date.now();
+miniget('https://api.github.com/repos/fent/node-ytdl-core/releases/latest', { headers: { 'User-Agent': 'a/b' } })
+  .text().then(response => {
+    if (!process.env.YTDL_NO_UPDATE && Date.now() - lastUpdateCheck >= 1000 * 60 * 60 * 12) {
```
fent, line 450: should go outside so that the request is not made until 12 hrs have passed

lib/info.js
```diff
@@ -443,6 +443,17 @@ for (let funcName of ['getBasicInfo', 'getInfo']) {
   };
 }
 
+const { version } = require('../package.json');
+let lastUpdateCheck = Date.now();
+miniget('https://api.github.com/repos/fent/node-ytdl-core/releases/latest', { headers: { 'User-Agent': 'a/b' } })
```
fent, line 448: put inside the function above, at line 440. otherwise, this is called only once when ytdl-core is first required. i also think this would all look cleaner if it was encapsulated in a function called `checkForUpdate` and then calling that function.

fent (changes requested):

lib/info.js
```diff
@@ -426,6 +426,20 @@ const getM3U8 = async(url, options) => {
   return formats;
 };
 
+// Check for updates.
+const { version } = require('../package.json');
+let lastUpdateCheck = Date.now();
+const checkForUpdates = () => {
+  if (!process.env.YTDL_NO_UPDATE && Date.now() - lastUpdateCheck >= 1000 * 60 * 60 * 12) {
+    miniget('https://api.github.com/repos/fent/node-ytdl-core/releases/latest', { headers: { 'User-Agent': 'a/b' } })
```
fent, line 434: i was looking at github's docs for their api
https://developer.github.com/v3/#user-agent-required

let's put "ytdl-core" here for the User-Agent. they're hosting the repo after all.

redbrain: this fails the linter bc the line's too long. what do you suggest i do?

fent: add line breaks between the options
```js
..., {
  headers: { 'User-Agent': 'ytdl-core' },
})
```

redbrain:
what is codecov, and is the error something i can resolve?

fent:
> what is codecov, and is the error something i can resolve?

it's for code coverage, as in is the how you introduced tested. if you're up for it, you can add tests. otherwise, i can do it after the merge.

and there are some eslint warnings again

redbrain:
Lint's throwing `comma-dangle` on the console.warn line, and I have no clue why. I'm assuming the rest of the errors stem from that.
As for testing, I'm not knowledgeable for writing tests, so I'd rather you work on that.
Do you know why the linter is screaming again? I can't figure that out.

fent:
you can run `npm run lint:fix` it should fix those lint errors for you

fent:
going to merge this as is, then correct the lint errors and add tests. thanks @redbrain! looking forward to the next update

meoyawn:
shouldn't it be part of https://github.com/fent/node-ytdl and not the core? People use core as a library in their projects

redbrain:
The goal is to alert library users. (it will alert node-ytdl users as well since it relies on the library)

meoyawn:
gotcha 😕

I'd use `npm` or `yarn` for that

redbrain:
> and there's https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/ and other bots

See the first message in this discussion for why this was implemented:

> upgrading the package to latest seems to be something many users forget to do, and they encounter errors because of it.

lib/info.js
```diff
@@ -426,6 +426,21 @@ const getM3U8 = async(url, options) => {
   return formats;
 };
 
+// Check for updates.
+const { version } = require('../package.json');
```
stoically, line 430: This breaks when bundling with webpack, not sure why. It imports the package.json from my project root, instead of ytdl-core's package.json. Workaround for me was to add a "version" field to my package.json, so it doesn't error out with 

```
TypeError: pkg.version is undefined
```

Might very well be me configuring webpack incorrectly, but thought I'd mention it here.

Repo to reproduce the issue: https://github.com/stoically/humanetube/

fent: weird. i'd ask webpack about it. otherwise, not sure what's causing webpack to import the wrong `package.json`

node's `require` works a bit different for relative paths, it looks for files relatively to the file it is required from.

redbrain: I bundled it with Browserify and it worked fine; I don't know enough about Webpack to help you further.

PR #792, opened 2020-11-27
Handle unicode encoded equals in ID token

benkaiser:
When running ytdl on a mobile device in [a chrome extension](https://github.com/benkaiser/Stretto-Helper-Extension) I'm working on, I was receiving ID_TOKEN with a unique encoded equals sign at the end.

The targeted fix here is to replace the individual bad character, but I'm open to other options. It may also be worth digging into why exactly this difference is triggered on mobile but not the desktop when running ytdl in the chrome extension.

### DIFF
--- a/lib/info.js
+++ b/lib/info.js
@@ -104,7 +104,7 @@ const getIdentityToken = (id, options, key, throwIfNotFound) =>
     if (!match && throwIfNotFound) {
       throw new UnrecoverableError('Cookie header used in request, but unable to find YouTube identity token');
     }
-    return match && match[2];
+    return match && match[2].replace('\\u003d', '=');
   });
 
 


lib/info.js
```diff
@@ -104,7 +104,7 @@ const getIdentityToken = (id, options, key, throwIfNotFound) =>
     if (!match && throwIfNotFound) {
       throw new UnrecoverableError('Cookie header used in request, but unable to find YouTube identity token');
     }
-    return match && match[2];
+    return match && match[2].replace('\\u003d', '=');
```
fent, line 107: does it need to be `/\u003d/g` to replace all, or does this character only occur once at the end?

fent (commented):
this seems like a strange phenomenon since it only occurs on mobile and only for that one character and only at the end. i'm inclined to take your word on it though, this is hard to test for.

lib/info.js
benkaiser: I'm honestly not sure, it only occurred once at the end in my case, so depends on if you'd like to keep this fix scoped as granularly as possible or not.

benkaiser:
@fent thanks, I agree it is quite strange and indeed hard to test for given it only showed up in that specific scenario.

I worked around it for now in my extension: https://github.com/benkaiser/Stretto-Helper-Extension/blob/cc4590b509d7813dcbb9fdc3d5efc9dddbda3c25/src/background.ts#L52

lib/info.js
fent: mind placing a comment above that explains this oddity on mobile?

benkaiser:
Sure enough, debugging on mobile it now is coming through correctly without the fix:

image

I'll close this PR for now as it may have been a transient issue.

Thanks for all your great work @fent

PR #803, opened 2020-12-02
Temp. fix age-restricted videos playback

modscleo4:
This change fixes age-restricted playback, as /embed/ status returns UNPLAYABLE instead of LOGIN_REQUIRED

A better solution is to transform UNPLAYABLE to LOGIN_REQUIRED on getEmbedPage, so really unavaillable videos don't use all 3 endpoints


fent:
ah, so age restricted videos are still available from the get_video_info endpoint. i must've tested incorrectly before.

fent:
thanks! going to commit a fix inspired from this PR

PR #805, opened 2020-12-03
fixed undefined for videoDetails.author.name

FishyOnMee:


Peppy8651:
Could you mention the bug or issue that this comes from? If it doesn't exist, then perhaps you could add the error in your description and also how it fixes the error?

Peppy8651:
Nevermind, I saw your commits and it looks like you're just fixing the example. However, info.videoDetails.author.name still works for me. Instead of just changing the example, perhaps you could add a comment using // to show that you can also use ownerChannelName if info.videoDetails.author.name is being buggy or returning undefined?

Peppy8651 (commented):
Maybe this could use some rewording, how about instead of "use the example below", you can add, "replace the line above with the line below and remove the '//'."

fent:
is there a video where the name is `undefined`

FishyOnMee:
nope, one day it stopped working. returned undefined and I started this convo: https://discord.com/channels/484464227067887645/484483415731601430/781207032342249542

fent:
the same field is used to get `author.name` btw

https://github.com/fent/node-ytdl-core/blob/52acff7d38cc45496c6c6a7b6b2eaaa6887c8aba/lib/info-extras.js#L115

if you're getting undefined, it could be an error

FishyOnMee:
alright then, I guess ill close this then.

PR #811, opened 2020-12-06
update youtube client version

dgivoli:
The old client version yielded error when fetching watch json url


dgivoli:
@fent I was trying to get adPlacements from the playResponse object, but the initial call to the "watch" url with "pbj=1" did not work because of a wrong "x-youtube-client-version" header in the request.
Looking youtube.com I noticed the client version has changed from "2.20201117.05.00" to "2.20201203.06.00"

fent:
thanks!

Issue #819, opened 2020-12-12
Stream doesn't end when video is done downloading


ryanburnette:
Environment info

```bash
$ node -v
v14.15.1

$ cat package.json | grep ytdl-core
    "ytdl-core": "^4.1.4"
```

Code I'm running: (the demo from readme)

```js
const fs = require('fs');
const ytdl = require('ytdl-core');
ytdl('http://www.youtube.com/watch?v=aqz-KE-bpKQ')
  .pipe(fs.createWriteStream('video.mp4'));
```

The issue is that the stream doesn't end at the end of the video, so the process doesn't exit.

ryanburnette:
This works though...

```js
const fs = require("fs");
const ytdl = require("ytdl-core");
ytdl("http://www.youtube.com/watch?v=aqz-KE-bpKQ")
  .pipe(fs.createWriteStream("video.mp4"))
  .end(function () {
    process.exit();
  });
```

ryanburnette:
This is what I needed. Just needed a way to know when the download is done.

```js
ytdl("http://www.youtube.com/watch?v=aqz-KE-bpKQ")
  .pipe(fs.createWriteStream("video.mp4"))
  .on("close", function () {
    process.exit();
  });
```

coolaj86:
That will sometimes result in the application being closed before the last packet of data is written to disk. Use this instead:

```js
let ws = fs.createWriteStream("video.mp4");
ytdl("http://www.youtube.com/watch?v=aqz-KE-bpKQ")
  .pipe(ws)
  .on("close", function () {
    ws.close();
  });
ws.on('close', function () {
    process.exit();
});
```

Pipes don't close automatically because... some reason I can't remember, but it's a valid reason. Pipes are used for more than just files - http streams, etc.

As to why the program just hangs there instead of exiting... probably missing a weak reference (`unref()`) somewhere on a timer or network connection:

- https://nodejs.org/api/net.html#net_server_unref
- https://nodejs.org/api/timers.html#timers_timeout_unref

ryanburnette:
> As to why the program just hangs there instead of exiting... probably missing a weak reference (unref()) somewhere on a timer or network connection:

Based on this, I'm going to reopen the issue so one of the maintainers can take a look.

fent:
the program hangs because the request cache uses timeouts

coolaj86:
> request cache

Where is that? Could we add some `.unref()`s to those timeouts to keep them weak?

[...]

PR #821, opened 2020-12-12
use weak timeout (rather than strong timeout)

coolaj86:
As per <https://nodejs.org/api/timers.html#timers_timeout_unref>, calling `.unref()` will mark the timer as weak in the event loop and then return the timeout object.

Fixes https://github.com/fent/node-ytdl-core/issues/819.


fent:
cool! i wasn't aware of this API on timeouts, pretty handy.

one request, do you mind removing the `process.exit()` line from `example/ffmpeg.js`. it's only there because of the timeouts hanging.

coolaj86:
@fent Done.

fent:
thanks!

PR #823, opened 2020-12-14
Fix Search Cache compatibility with Electron Apps

PrestonN:
The recent change made in v4.1.5 breaks compatibility with Electron based applications. This is because the renderer process within Electron uses a similar library as the one within Node JS, but has small differences to the actual library being used. You can read more about it (and the suggested fix) over at [electron/electron#21162](https://github.com/electron/electron/issues/21162).

This change uses the suggested line in the linked issue to specify that `setTimeout` should use the functionality built into Node instead of the Electron based one.

I've already tested this fix in my project and it seems to work fine with this change.

### DIFF
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -1,3 +1,5 @@
+const { setTimeout } = require('timers');
+
 // A cache that expires.
 module.exports = class Cache extends Map {
   constructor(timeout = 1000) {
--- a/lib/info.js
+++ b/lib/info.js
@@ -3,6 +3,7 @@ const querystring = require('querystring');
 const sax = require('sax');
 const miniget = require('miniget');
 const utils = require('./utils');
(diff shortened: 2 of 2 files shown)


fent:
til something new about electron.

thanks for the fix! do you mind adding the same import on `info.js`? it also uses `setTimeout`, and although it doesn't use `.ref()`, i'd like to keep them consistent

PrestonN:
Done!

I actually wasn't aware of this either, so it was weird to see once I noticed the error and started looking up the reason for it.

Everything works without adding the same line to `info.js` however I went ahead and added it per your request. I also left a comment there for future reference.

fent:
thanks!

meoyawn:
thanks! this also fixes compat with Cloudflare Workers

PR #832, opened 2020-12-16
feat: expose version

TimeForANinja:
the package.json of a submodule is hard to reach
that's all i'll say

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,10 @@ Throws an Error if it fails to parse an ID.
 Same as the above `ytdl.getURLVideoID()`, but can be called with the video ID directly, in which case it returns it. This is what ytdl uses internally.
 Throws an Error if it fails to parse an ID.
 
+### ytdl.version
+
+The version string taken directly from the package.json.
+
 ## Limitations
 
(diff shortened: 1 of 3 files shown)


fent:
if this is for the auto-updater, i don't think exposing `version` can help with arbitrary modules. i was looking at the node modules API, there's `Module._resolveFilename`(not documented, so may be unavailable in the future), and `new Module(Module._resolveFilename('ytdl-core')).path` which returns the absolute path of a required file.

after you have the path, you can use something like https://www.npmjs.com/package/find-root

TimeForANinja:
that's basically my current approach - just without any dependencies and using `require.cache[require.resolve(xxx))` instead of `Module_resolveFilename` which i'm unable to find

that updater has some bigger design challenges with updating sub-dependencies, removing deleted files, handling non-github-repositories and so on....

exposing version was just a nice-to-have for some minor projects i run
not forcing everyone to use complex methods for reaching the package.json

fent:
where do we stand on this?

TimeForANinja:
🤷🏿 i'm still in favor as a nice-to-have
but i don't mind not adding it

PR #835, opened 2020-12-18
docs: add manually-generated docs for use with GitHub Pages

redbrain:
Hey there fent,
Some proper documentation would do ytdl-core a lot of good. I tried a couple times (which you saw) and now it's in PR form.
***
I made the docs from the `typings/index.d.ts` file into `/docs` with Typedoc 0.19. Here's how I got it to work with only a .d.ts file:
```
typedoc --out ./docs --includeDeclarations --excludeExternals --entryPoint '"ytdl-core".ytdl' --readme none --mode file --theme minimal typings/index.d.ts
```
Given that Typedoc will be releasing a major update soon, and that you'll hopefully port the actual library to Typescript, _I wouldn't recommend automating this process until both of those things happen._ Until then, I'm happy to generate these docs for you whenever a new version of ytdl-core is released with new typings.
***
These are the settings I used in GitHub Pages to deploy the docs site:
alt: shows the site being built from the `/docs` folder in the `master` branch
You can see what the docs site looks like on my fork: https://redbrain.github.io/node-ytdl-core/
***
If merged and set up the same way, the docs would be located at `https://fent.github.io/node-ytdl-core/`.
You could also set the Pages folder to `/`; then the above address has the README, and then `...ytdl-core/docs/` has the docs.
***
That's all the info/suggestions I have for you; I'm open to any info/suggestions that you have for this.
You and other maintainers have edit permissions here, so if I made a simple mistake anywhere go ahead and correct it.
Thanks for your hard work on the library; it's a blessing to me and others. Stay safe!


redbrain:
We should probably exclude the docs folder from the linter; however I don't know how to do that.

fent:
hey redbrain, thanks for looking into this. is there any way to add annotations to each of the fields from the .dts file? like maybe adding comments above each fields would put that in the generated html?

if the output for this is html, i'd rather keep it in a separate branch. it would keep the master branch less cluttered. if someone sees the `/docs` folder and goes to it (https://github.com/redbrain/node-ytdl-core/blob/master/docs/index.html) they'd see raw unrendered html.

what do you think?

redbrain:
Comments above definitions will be added to the rendered HTML; see typedoc.org/guides/doccomments .
Since much of your actual code's already commented, waiting for full TypeScript becomes even more appealing.
If you'd like me to continue with adding comments or moving to another branch, let me know; or, I can also close the PR until full TypeScript arrives

BR88C:
I feel that it would also be helpful to append some form of a guide to the docs (have the suggested docs above along with a guide section) as a lot of common questions could be answered there and some people may have trouble at first with understanding the docs suggested above, it wouldn't need to be super complex and it could be easily done by using something simple like docsify to make changes easier as it's just markdown files.

fent:
> Since much of your actual code's already commented, waiting for full TypeScript becomes even more appealing.
> If you'd like me to continue with adding comments or moving to another branch

adding comments to current `index.d.ts` definitions that are already in `lib/` seems like double work, i'd rather we wait for ts. if you'd like to add comments to responses there, as opposed to exposed ytdl functions, then those could be used during the conversion to ts.

ideally i'd like to keep everything in the repo, and have the docs be markdown. if there's no tool for that, we can ultimately do html.



> I feel that it would also be helpful to append some form of a guide to the docs (have the suggested docs above along with a guide section) as a lot of common questions could be answered there and some people may have trouble at first with understanding the docs suggested above, it wouldn't need to be super complex and it could be easily done by using something simple like docsify to make changes easier as it's just markdown files.

that's what the `example/` folder was meant to do. the examples are very general and small though. there's discord bot for example, but it's very simplistic.

if there's anything in the readme that's unclear or confusing, please let me know. just from looking at it now, i think we could 
- remove the various ways to import ytdl-core, as that's more related to the environment the user is on, and not related to ytdl-core
- have an example after every exported method to give an idea on how it's used
- link to the `examples/` folder at the top after the first example to give it more exposure

BR88C:
> ideally i'd like to keep everything in the repo, and have the docs be markdown. if there's no tool for that, we can ultimately do html.
> 
> >
That's why I suggested [docsify](https://docsify.js.org/#/), it's on the fly rendering of markdown, so people could either look at markdown files in github via navigating through folders or in github pages. All we would need to do is have github pages point to the main html that's needed for docsify, then that html just grabs the markdown files and does the rest of the rendering for you

>that's what the example/ folder was meant to do. the examples are very general and small though. there's discord bot for example, but it's very simplistic.
>
> >

Yeah I agree, and like you said we should definitely point to them in the README. Those files also should probably have more comments as the point is to show people how the code works, not to just provide something to copy paste. However I still think we need some form of an FAQ as there are so many repeated tickets, questions in Discord, etc, and compiling them in an easily accessible and understandable manor would benefit a lot of people and reduce the number of tickets on problems that already have solutions

TimeForANinja:
@redbrain https://eslint.org/docs/user-guide/configuring#ignoring-files-and-directories
should be easy since we already have a eslint config file 😉

redbrain:
Docsify seems to just create webpages from existing Markdown.
Since almost all of ytdl-core's documentation is in README, that doesn't appear to be necessary.
I had the goal of creating docs as a reference for navigating the varied info.json structure,
which was why it made sense to do it automatically from typings.
While I'd love to make an intuitive documentation/guide thing for YTDL, I do not have enough time to do that.

redbrain:
I have an idea for creating typing docs that are mostly human-readable in source format.
You can ask Typedoc to output in JSON format instead of HTML. Replacing `--out ./docs` with `--json ./docs.json` yields this:
https://hastebin.com/ebosixibad.json
Still lengthy and not the nicest to stare at, but you can probably turn this into something readable with some JS magic.
Since it's now an object, you could automate the creation of bulleted Markdown, which is obviously readable in source format.
If anyone wants to jump on that you can, otherwise I would wait until the project's TS and put rendered HTML in another branch.

redbrain:
Closing this as it's low-priority and basically unneeded for use.

talks2much:
Why don't we just replace API section with link to this auto generated docs: https://paka.dev/npm/ytdl-core?

TimeForANinja:
> Why don't we just replace API section with link to this auto generated docs: https://paka.dev/npm/ytdl-core?

That website just showcases the typescript typings. Those are handwritten and might include errors.
Sth. like that would be a possibility after a full typescript port - adding in-code descriptions of the functions & so on.

I guess for now we'll stay with the API-Section sine it's also a commitment to keep the api stable over releases

PR #850, opened 2020-12-31
Add rich thumbnail (animated thumbnail) to related videos

Aasim-A:
With a bit of digging, I found out that YouTube only returns the animated thumbnails on supported browser through the `user-agent` header.
Added tests to cover the code.


test/basic-info-test.js
```diff
@@ -83,6 +83,44 @@ describe('ytdl.getBasicInfo()', () => {
...
+          requestOptions: {
+            headers: {
+              'User-Agent': 'yay',
+            },
+          },
+        });
+        scope.done();
+        assert.deepEqual(info.related_videos[0].richThumbnails, []);
+      });
+    });
```
fent, line 123: no need to add these. we only need the tests in `info-extras-test.js`

lib/info.js
```diff
@@ -43,6 +43,14 @@ const AGE_RESTRICTED_URLS = [
 */
 exports.getBasicInfo = async(id, options) => {
   const retryOptions = Object.assign({}, miniget.defaultOptions, options.requestOptions);
+  if (options.requestOptions === undefined) {
+    options.requestOptions = {};
+  }
```
fent, line 48: we should try not to alter any objects passed to the function. this will add a `requestOptions` to the original `options`, which may confuse users who later see an extra field. we can instead overwrite the `options` given with one that contains a default `requestOptions`
```suggestion
options = Object.assign({}, options, {
  requestOptions: {},
});
```

test/info-extras-test.js
```diff
@@ -175,6 +176,13 @@ describe('extras.getRelatedVideos()', () => {
     assertRelatedVideos(extras.getRelatedVideos(info));
   });
 
+  describe('With richThumbnails', () => {
+    it('Returns related videos', () => {
+      const info = require('./files/videos/rich-thumbnails/expected_info.json');
+      assertRelatedVideos(extras.getRelatedVideos(info));
```
fent, line 182: pass an extra param here to check that `richThumbnails` is not empty

fent (changes requested):
hi, thank you for the PR 👍 

i didn't know adding a good user-agent would provide additional fields, do you know if this affects only the watch.json endpoint, or does it affect the watch.html page too? i noticed that only `watch.json` is in the `videos/rich-thumbnails` folder, is why i ask. also, add `rich-thumbnails` to `test/files/refresh.js` so others can refresh those files in the future if needed.

i wonder if there are more fields that we haven't noticed? i did suspect that the watch.json endpoint is slightly more reliable with a user agent, but couldn't confirm when testing.

Aasim-A:
>do you know if this affects only the watch.json endpoint, or does it affect the watch.html page too

Yes, it affects html too, the same `richThumbnails` will be returned, should I add tests on the html response?

test/basic-info-test.js
Aasim-A: This is added to test the header override

lib/info.js
Aasim-A: this actually breaks a couple of tests, I got it working with:
```javascript
options.requestOptions = Object.assign({}, options.requestOptions, {});
```

test/basic-info-test.js
fent: ah, if this is checking if users can still provide a custom `user-agent`, the test should be renamed, and should not test `richThumbnails`. leave the `richThumbnails` test in `info-extras-test`

fent:
> Yes, it affects html too, the same `richThumbnails` will be returned, should I add tests on the html response?

one is fine

test/info-extras-test.js
fent: i meant here to call it like `assertRelatedVideos(extras.getRelatedVideos(info), true)` and inside the function make sure the `richThumbnails` array for each video is not empty, since it shouldn't be empty for this test.

Aasim-A: Oh, ok

Aasim-A: done.

fent (approved):
awesome! thanks again and happy new year!

PR #851, opened 2021-01-01
Add `storyboards` to `info.videoDetails`

Aasim-A:
The `storyboards` `spec` field has the information of the available storyboards encoded inside.

### DIFF
--- a/lib/info-extras.js
+++ b/lib/info-extras.js
@@ -268,3 +268,54 @@ exports.cleanVideoDetails = videoDetails => {
     'videoDetails.shortDescription', 'videoDetails.description');
   return videoDetails;
 };
+
+/**
+ * Get storyboards info.
+ *
+ * @param {Object} info
+ * @returns {Array.<Object>}
+ */
+exports.getStoryboards = info => {
+  const parts = info.player_response.storyboards &&
+    info.player_response.storyboards.playerStoryboardSpecRenderer &&
+    info.player_response.storyboards.playerStoryboardSpecRenderer.spec &&
+    info.player_response.storyboards.playerStoryboardSpecRenderer.spec.split('|');
+
+  if (!parts) return [];
+
+  const url = new URL(parts.shift());
+
+  return parts.map((part, i) => {
+    let [
+      width,
+      height,
+      count,
+      storyboardWidth,
+      storyboardHeight,
+      interval,
+      nameReplacement,
+      sigh,
+    ] = part.split('#');
+
+    url.searchParams.set('sigh', sigh);
+
+    count = parseInt(count, 10);
+    storyboardWidth = parseInt(storyboardWidth, 10);
+    storyboardHeight = parseInt(storyboardHeight, 10);
+
+    const storyboardCount = Math.ceil(
+      count / (storyboardWidth * storyboardHeight),
+    );
+
+    return {
+      templateUrl: url.toString().replace('$L', i).replace('$N', nameReplacement),
+      width: parseInt(width, 10),
+      height: parseInt(height, 10),
+      count,
+      interval: parseInt(interval, 10),
+      storyboardWidth,
+      storyboardHeight,
+      storyboardCount,
+    };
+  });
+};
--- a/lib/info-extras.js
+++ b/lib/info-extras.js
@@ -287,9 +287,9 @@ exports.getStoryboards = info => {
 
   return parts.map((part, i) => {
     let [
-      width,
-      height,
-      count,
+      thumbnailWidth,
+      thumbnailHeight,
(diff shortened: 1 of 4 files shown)


test/info-extras-test.js
```diff
@@ -244,3 +244,33 @@ describe('extras.getDislikes()', () => {
...
+describe('extras.getStoryboards()', () => {
+  it('Returns storyboards', () => {
+    const info = infoFromWatchJSON('no-likes-or-dislikes');
+    const storyboards = extras.getStoryboards(info);
+
+    assert.ok(Array.isArray(storyboards));
+    assert.ok(storyboards.length > 0);
+
+    for (let storyboard of storyboards) {
+      assert.ok(storyboard.templateUrl);
```
fent, line 257: use `assetURL()`

fent (commented):
you're on a roll 👍 

i'm not familiar with youtube storyboards. what do they represent? is it the cards that pop up near the end of videos or something else? could you provide an example video that has them?

Aasim-A:
Sure, the storyboard is a preview for the video represented as pictures, you see them above the seeking bar when seeking the video.

Youtube storyboard

lib/info-extras.js
```diff
@@ -268,3 +268,54 @@ exports.cleanVideoDetails = videoDetails => {
...
+
+  const url = new URL(parts.shift());
+
+  return parts.map((part, i) => {
+    let [
+      width,
+      height,
+      count,
+      storyboardWidth,
+      storyboardHeight,
```
fent, line 294: what is `width` and `height` compared to `storyboardWidth`, `storyboardHeight`?

lib/info-extras.js
```diff
@@ -268,3 +268,54 @@ exports.cleanVideoDetails = videoDetails => {
...
+      interval,
```
fent, line 295: i'm seeing interval be `0` for some, is that expected?

lib/info-extras.js
```diff
@@ -268,3 +268,54 @@ exports.cleanVideoDetails = videoDetails => {
...
+      nameReplacement,
```
fent, line 296: name replacement can be "M$M", but when that is added to the url, the url gives a 404. i replaced it with "M0" (by looking at the network tab) and that worked

fent (changes requested):

lib/info-extras.js
Aasim-A: Basically when you open the url you'll see a big picture comprised of small pictures representing different times of the video, the `width` and `height` are the dimensions of the small picture, and  `storyboardWidth` and `storyboardHeight` are the dimensions of the storyboard

lib/info-extras.js
Aasim-A: Yes, this is the default storyboard, the lowest quality and the small thumbnails represent bigger time spans, this should be used if the others are not available or if the network is slow to show something instread of blank

lib/info-extras.js
Aasim-A: hints the name `templateUrl`, it's meant to edited before request, changing `$M` to the index of the story board based on `storyboardCount`. And BTW these fields are provided by YouTube, not me 😉

Aasim-A: Adding an example would clearify the use of storyboards

lib/info-extras.js
fent: how about `rows` and `columns`?

lib/info-extras.js
fent: ah i see. yep that wasn't too clear to me when i was looking at it

lib/info-extras.js
Aasim-A: > how about `rows` and `columns`?

What will they represent? and will it replace other fields? or just an addition

fent: i mean in place of `storyboardWidth` and `storyboardHeight`. the current names can imply pixel width not images per side

Aasim-A: Oh I thought these fields would give pixel count, yes rows and columns would be much clearer

fent: and what do you think of instead of `width`/`height`, using `thumbnailWidth`/`thumbnailHeight`? i don't feel as strongly about this change as `rows`/`columns`, so would like to know what you think

Aasim-A: Hmmm, it would be more easier to read though, I think we should change it

fent (approved):
looks good! thanks again for taking the time in figuring out how this works and filing a PR 🎉

Aasim-A:
No worries! 🙂

PR #848, opened 2020-12-23
feat: add `nextVideo`

Alexandre94H:
To get the next video from Youtube autoplay

closes #810

PR #853, opened 2021-01-03
fix: empty related videos when able to choose the topic of related videos

### LINKED ISSUE #810: new feature: next video
Can someone add parsedBody.response.contents.twoColumnWatchNextResults.autoplay.autoplay.sets[0].autoplayVideo.watchEndpoint.videoId to have the next video ?

skick1234:
image

Fix empty related videos when the watch page contains autoplay feature

It happens in some regions or when using cookies.

closes #810 
closes #848


skick1234:
`info.nextVideo = info.related_videos[0]` can close #848 too. But I think it's not neccesary.

fent (changes requested):
hi, thanks for another PR

how did you generate the `watch.html` page with autoplay on? is that something that can be turned by a cookie or user setting? i'd like to add it to `test/files/refresh.js` in case it needs to be updated in the future

skick1234:
Actually, this is not because the autoplay feature is on.

When I logged in, related videos can be sort (e.g. All, Cats, Recently Uploaded). This makes related videos empty.

My autoplay json file
image
image

I tried to remove all the account-login cookies (keep PREF, ... cookies) but I cannot reproduce that. Maybe this is a user-only feature or a beta feature, idk.

regular json file
image
image

skick1234:
Idk there is a way to shorten this case :>

fent:
hmm, so the saved html and json pages can't be added to `refresh.js` since the new format won't be on there? are the json screenshots from requests in your browser?

skick1234:
> hmm, so the saved html and json pages can't be added to `refresh.js` since the new format won't be on there?

I'm afraid that can't be done without an user cookie.

> are the json screenshots from requests in your browser?

I logged from miniget response with my clone's cookie.

skick1234 (commented):
Reviewed

fent (commented):
i still think `related-topics` should be added to `refresh.js`, only for future maintenance. just add a note that it needs a `requestOptions: { headers: 'YOUR_COOKIE' }`.

btw, the `watch.html` file contains possibly identifying values from the account you used, things like `ID_TOKEN` and `INNERTUBE_API_TOKEN`. i'm actually not sure how compromising these are, but could be worth removing them if you have concerns. or maybe you used a throwaway account and it doesn't matter

skick1234:
I used my clone account to run `refresh.js`.

Maybe we should add a cookie to environment variables. Or just update it manually :>

fent (approved):
thank you!

PR #855, opened 2021-01-08
docs(typings): more `VideoInfo` type mappings

Aasim-A:
Added missing types, corrected a couple of incorrect types

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -108,13 +108,23 @@ declare module 'ytdl-core' {
       isTranslatable: boolean;
     }
 
+    interface audioTrack {
+      captionTrackIndices: number[];
+    }
+
+    interface translationLanguage {
+      languageCode: captionTrack['languageCode'];
+      languageName: captionTrack['name'];
+    }
+
     interface VideoDetails {
       videoId: string;
       title: string;
       shortDescription: string;
       lengthSeconds: string;
-      keywords: string[];
+      keywords?: string[];
       channelId: string;
+      isOwnerViewing: boolean;
       isCrawlable: boolean;
       thumbnail: {
         thumbnails: thumbnail[];
@@ -174,7 +184,7 @@ declare module 'ytdl-core' {
       };
       lengthSeconds: string;
       ownerProfileUrl: string;
-      ownerGplusProfileUrl: string;
+      ownerGplusProfileUrl?: string;
       externalChannelId: string;
       isFamilySafe: boolean;
       availableCountries: string[];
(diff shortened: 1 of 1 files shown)


typings/index.d.ts
```diff
@@ -206,12 +216,13 @@ declare module 'ytdl-core' {
...
       age_restricted: boolean;
-      likes?: number;
-      dislikes?: number;
+      likes: number | null;
+      dislikes: number | null;
       media: Media;
       author: Author;
       thumbnails: thumbnail[];
       storyboards: storyboard[];
+      description: string | null;
```
fent, line 225: this goes in `MoreVideoDetails`

fent (changes requested):

typings/index.d.ts
fent: it's also not `| null`, is it? the property is based on `shortDescription: string`

Aasim-A: After testing, if there's no description, the `description` field is `null`.

Aasim-A: > this goes in `MoreVideoDetails`

It's already there

fent: it is? i don't see it. this property is retrieved from the raw `videoDetails` btw

https://github.com/fent/node-ytdl-core/blob/master/lib/info-extras.js#L265

Aasim-A: It being `null` makes sense, 
```js
videoDetails.description = videoDetails.shortDescription || getText(videoDetails.description);
```
when there's no description, `shortDescription` is `""`, and that is a falsy value, so it assigns whatever is returned from the second part into `videoDetails.description`, if there's no description there isn't anything other than `shortDescription` in the `info`, so the function `getText` returns null in this case.
The expression `??` solves this problem because it moves to the next part if the first is `null` or `undefined`, but that would require node 14 at least.

Aasim-A: We could leave it like this or change the assignment to return empty string, what do you think?
sth like this:
```js
videoDetails.description = videoDetails.shortDescription || getText(videoDetails.description) || '';
```

fent: i agree with it being `null`. but i still don't see how it's available in the raw `VideoDetails`. it's in `MicroformatRenderer`, and then in `MoreVideoDetails` after being added from the above line

Aasim-A: I'm not sure if I understand you correctly, as far as I can see, it's within `MoreVideoDetails`, unless you're talking about something else
Capture

fent: oh! my mistake. i didn't look at the diff correctly and thought it was below `VideoDetails`. sorry

fent (approved):
thanks!

PR #862, opened 2021-01-20
fix escaped backslashes towards end of string & add test

### LINKED ISSUE #861: cutAfterJSON failing with escaped backslashes
stumbled into some problems with ytsr
narrowed it down to the cutAfterJSON function, which is based on ytdl-core's utils#cutAfterJSON function

the function only does a simple lookup if the previous character is a backslash when it checks for the end of strings
if e.g. the channelname ends with a backslash that translates to `"xxx\\"` which lets the cutAfterJSON function fail

**low priority issue**
tried the ytdl-core#getInfo on one of the channels videos and it succeeded
guess the function is not always used to retrieve the json

problematic line: https://github.com/fent/node-ytdl-core/blob/master/lib/utils.js#L81

code to reproduce:
```js
require('./lib/utils.js').cutAfterJSON('{"text": "\\\\"};');
```

opening this issue instead of a pr since i feel like using a regex like `[^\\]\\(\\\\)*` or even a backwards loop suck performance wise and i am unable to come up with sth better

Channel that Errors: https://www.youtube.com/channel/UC7Dk-eKv2KB7EO28UsFzqdg
Related issue: https://github.com/TimeForANinja/node-ytsr/issues/98

TimeForANinja:
instead of nested loops or some lookback regex i decided on adding a simple boolean to track if the current character is escaped by a previous backslash or not

also added a test case to check for this in the future

closes #861

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -88,8 +88,7 @@ exports.cutAfterJSON = mixedJson => {
 
     // Toggle the isEscaped boolean for every backslash
     // Reset for every regular character
-    if (mixedJson[i] === '\\') isEscaped = !isEscaped;
-    else isEscaped = false;
+isEscaped = mixedJson[i] === '\\' && !isEscaped;
 
     if (isString) continue;
 
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -72,16 +72,25 @@ exports.cutAfterJSON = mixedJson => {
   // States if the loop is currently in a string
   let isString = false;
 
+  // States if the current character is treated as escaped or not
+  let isEscaped = false;
+
   // Current open brackets to be closed
   let counter = 0;
 
   let i;
   for (i = 0; i < mixedJson.length; i++) {
     // Toggle the isString boolean when leaving/entering string
-    if (mixedJson[i] === '"' && mixedJson[i - 1] !== '\\') {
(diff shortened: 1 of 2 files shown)


lib/utils.js
```diff
@@ -72,16 +72,25 @@ exports.cutAfterJSON = mixedJson => {
...
-    if (mixedJson[i] === '"' && mixedJson[i - 1] !== '\\') {
+    if (mixedJson[i] === '"' && !isEscaped) {
       isString = !isString;
       continue;
     }
+
+    // Toggle the isEscaped boolean for every backslash
+    // Reset for every regular character
+    if (mixedJson[i] === '\\') isEscaped = !isEscaped;
+    else isEscaped = false;
```
fent, line 92: does this work?

```suggestion
isEscaped = mixedJson[i] === '\\' && !isEscaped;
```

fent (commented):
hi, thanks for taking the time to fix this. haven't had much time to work on open source the past couple of weeks >_<

TimeForANinja:
> hi, thanks for taking the time to fix this. haven't had much time to work on open source the past couple of weeks >_<

all fine
nothing to important
and shows the fun part about open source - others doing your work 😉

Issue #882, opened 2021-02-19
No longer works in browserify after "refactor: use `URL` object to parse urls"


martinbrennan:
Until recently I have been able to Browserify ytdl-core so that it works in a browser
(I patch miniget to use a proxy mechanism to make fetches from Youtube or else they would be blocked by CORS)

The refactor: use `URL` object to parse urls change broke things because URL is called Url when require ('url') is invoked in a browser (as opposed to node) environment. Also the browser doesnt have a searchParams property.

I have patched these things and it works again in the browser.

I know this is not an anticipated use case - but I thought you might like to know.

Martin

fent:
since browserify is supposed to replicate node behavior, this looks a  browserify issue. but i'm willing to add workarounds to have ytdl-core work in various environments. what does `require('url')` do in browserify? do they not implement it? or is it different?

have you considered using rollup instead? i've h

martinbrennan:
Hi
Thanks for the quick answer - I suspect you are right that it is strictly a Browserify issue.
It appears that the constructor URL is called Url in the browser
I am not an Javascript programmer so my fixes are likely suboptimal
Generally I replaced
const { URL } = require('url');
with
const { Url } = require('url);
const URL = Url
To fix that problem.
The second problem was that searchParams is not available in the url library in the browser so I reverted the function setDownloadURL in sig,js to the code used before. It uses url.parse.
Hope thats clear enough.
Thank you for the rollup.js reference. I did not know about it. I will give it a try - though that may not happen until the next Youtube change triggers a rebuild
Martin

martinbrennan:
Hello again
Thanks again for the rollup.js suggestion.
I can confirm that rollup builds a bundle that works in the browser - so the problem really is for Browserify to fix - I am happy to use rollup.
Martin

fent:
cool! well then, we can keep this issue open to inform those using browserify that it is a browserify issue

vegidio:
For the record, since the OP edited his last message and this might not be clear to everyone that comes here, this is **NOT** an issue with Browserify alone.

I also tried to create a bundle to work in the browser using Browserify, Rollup, Webpack and Parcel. All of them failed with the same error `URL is not a constructor`.

fent:
@redbrain you had a similar issue on https://github.com/fent/node-ytdl-core/issues/864, did you get a reply on the polyfill repo?

redbrain:
[...]

PR #892, opened 2021-03-06
fix: Use URL from window if possible

jamezrin:
closes https://github.com/fent/node-ytdl-core/issues/882


redbrain:
LGTM! Hopefully we can push this along with the other two PRs mentioned in #882

TimeForANinja:
Hey
i removed all occurances of URL in #928 
Still thanks for the PR 😉

PR #895, opened 2021-03-07
Update README.md

Huxisher:
Add another cli wrapper base on node-ytdl but for spotify


Huxisher:
Add another cli wrapper base on ytdl-core but for spotify

PR #901, opened 2021-03-14
feat: Access to the various getInfo miniget requests.

gatecrasher777:
In my use case it is to monitor compressed/uncompressed bandwidth, but there could be many other uses (i.e. applying a timeout or request logging)

See discussion: https://github.com/fent/node-ytdl-core/discussions/899

I'm suggesting a callback which would receive each request in the getInfo pipeline, including retries.

Perhaps there is a better way... but this is my solution which is working well on my test app.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ Gets metainfo from a video. Includes additional formats, and ready to download d
 `options` can have the following
 
 * `requestOptions` - Anything to merge into the request options which [miniget](https://github.com/fent/node-miniget) is called with, such as `headers`.
-* `reqCallback` - Provide a callback function that receives miniget request stream objects used while fetching metainfo.
+* `requestCallback` - Provide a callback function that receives miniget request stream objects used while fetching metainfo.
 * `lang` - The 2 character symbol of a language. Default is `en`.
 
 ### ytdl.downloadFromInfo(info, options)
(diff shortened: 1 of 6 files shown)


gatecrasher777:
Good points. Will revise.

gatecrasher777:
Lint does not like == wants ===. lol.

TimeForANinja:
First of all thanks for contributing to this library
I wanna wait for fent to state his opinion on this thou
Besides that: Having a testcase would be cool 😅

gatecrasher777:
@TimeForANinja  
Thanks for the feedback.  By testcase, is that something I should provide in this PR?  Is it a test using requestCallback that you want to see added to the test folder?  Or is it an example using requestCallback that can be added to the example folder?  I already wrote a compressed/uncompressed bandwidth example that runs the 4 test cases that are used in full-info-test.js.  I would just need to merge it with this PR.

TimeForANinja:
> Is it a test using requestCallback that you want to see added to the test folder? 

That one.
If you check the `7 succesful checks` you can see, that we try to track the coverage of the module.
This helps us test the code we write.
When an Error appears for the first time we try to add a test case so that when we modify the code years later it checks that we don't bring back old bugs.
In this case it is not about previous errors but checking with every new version, that the callback is still used.

But, before you but that much work into this, i'd like to wait for fent to state his opinion.

TimeForANinja:
suggested some changes in https://github.com/gatecrasher777/node-ytdl-core/pull/1 😉

TimeForANinja:
thanks again for the pr 👍

PR #903, opened 2021-03-18
fix: Handle HLS audio only request edge case

### LINKED ISSUE #886: Livetreams only work for a few seconds with audioonly filter
When playing a live stream it only works for 2-3 seconds before stopping when using the `audioonly` filter.
Here is the code I am using:
```
let songInfo = ytdl.getInfo(link_of_lofi);
let connection = await msg.member.voice.channel.join();
let dispatcher = connection.play(ytdl.downloadFromInfo(songInfo, { highWaterMark: 1<<12, filter: 'audioonly' }));
```
Livestream example: https://www.youtube.com/watch?v=5qap5aO4i9A

Svallinn:
Closes #886

When providing a live stream video ID and audio only related options to either `quality` or `filter`, the library tends to pick out the only format with audio without video that exists, which tends to be a 5 second long stream.

YouTube seems to be using this stream in their live videos, but it has to re-request it with different parameters every few seconds.

I'm not entirely sure how the mechanism behind it works (it probably just needs to merge the segments in a certain way) or if the library will ever implement that, so for now, this is good way to not return that usually useless format to the user.

~EDIT: I haven't written tests, but I have manually tested it fully.~
EDIT2: Tests have now been written (I don't usually write tests though, so let me know if I need to change anything)

### DIFF
--- a/lib/format-utils.js
+++ b/lib/format-utils.js
@@ -121,16 +121,18 @@ exports.chooseFormat = (formats, options) => {
       format = formats[formats.length - 1];
       break;
 
-    case 'highestaudio':
+    case 'highestaudio': {
       formats = exports.filterFormats(formats, 'audio');
       formats.sort(sortFormatsByAudio);
-      // The formats video quality decreases with rank
-      // So we just need to skip forward to the last format the with same audio quality
-      // To get the format with highest audio + lowest video
-      do {
-        format = formats.shift();
-      } while (format && sortFormatsByAudio(format, formats[0]) === 0);
+      // Filter for only the best audio format
+      const bestAudioFormat = formats[0];
+      formats = formats.filter(f => sortFormatsByAudio(bestAudioFormat, f) === 0);
+      // Check for the worst video quality for the best audio quality and pick according
+      // This does not loose default sorting of video encoding and bitrate
+      const worstVideoQuality = formats.map(f => parseInt(f.qualityLabel) || 0).sort((a, b) => a - b)[0];
+      format = formats.find(f => (parseInt(f.qualityLabel) || 0) === worstVideoQuality);
       break;
+    }
 
     case 'lowestaudio':
       formats = exports.filterFormats(formats, 'audio');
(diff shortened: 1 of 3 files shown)


redbrain:
LGTM!

lib/format-utils.js
```diff
@@ -100,11 +106,40 @@ exports.chooseFormat = (formats, options) => {
...
+    const testAudioFilter = options.filter &&
+      typeof options.filter === 'string' &&
+      /audio(only)?/.test(options.filter);
+
+    if (testAudioQuality || testAudioFilter) {
+      formats = formats.filter(fmt => fmt.isHLS);
+      if (testAudioQuality) {
+        switch (options.quality) {
+          case 'highestaudio':
+            formats.sort(sortFormatsByHighAudioAndLowVideo);
```
TimeForANinja, line 123: is this not also needed for 'lowestaudio' and therefor all 3 cases?

Svallinn: There's no need.
I'm already filtering out all formats that aren't HLS from the array, so the lowest audio is automatically in the end.

`sortFormatsByHighAudioAndLowVideo` exists because, typically, some resolutions have the same audio bitrate.
p.e. it makes it so that if both `720p` and `1080p`exist (these typically have the same audio bitrate), it'll sort `720p` to the top of the array and choose it in order to reduce bandwidth use for the user.

TimeForANinja:
I am sorry but this still doesn't cut it for me 😅 
Lemme make a PR to suggest some changes

TimeForANinja:
😉 wrote my ideas in https://github.com/Svallinn/node-ytdl-core/pull/1

TimeForANinja:
@redbrain wanna have a look again?

redbrain:
will do

TimeForANinja:
thanks again for the pr 👍

joek85:
> but it has to re-request it with different parameters every few seconds

this is not true, if you get the URL and open it in the browser it plays for 5 seconds, if you hit refresh again with the same URL it plays again another 5 seconds.
so we have to find another approach to this, like feed the URL to some streamer and pipe the audio. or to send request every 2s for example and fill a buffer with the new audio data.
any suggestions would be appreciated.

TimeForANinja:
@joek85 pls open a new issue for the discussion 😉
This is a fix to filter out (currently) unplayable formats
Supporting those formats would be a new feature

PR #938, opened 2021-06-19
Add parameters to `get_video_info` endpoint to fix 404 errors

### LINKED ISSUE #939: MinigetError: input stream: Status code: 404
MinigetError: input stream: Status code: 404
    at ClientRequest.<anonymous>
    at Object.onceWrapper (events.js:422:26)
    at ClientRequest.emit (events.js:315:20)
    at HTTPParser.parserOnIncomingClient (_http_client.js:641:27)
    at HTTPParser.parserOnHeadersComplete (_http_common.js:126:17)
    at TLSSocket.socketOnData (_http_client.js:509:22)
    at TLSSocket.emit (events.js:315:20)
    at addChunk (internal/streams/readable.js:309:12)
    at readableAddChunk (internal/streams/readable.js:284:9)
    at TLSSocket.Readable.push (internal/streams/readable.js:223:10) {
  statusCode: 404
}
I have this issue evry time i try to play  a yt song ( mp3 file work and soundcloud songs work too) . I have this issue today and I havn't touch the code ; only show this error in the console

D3SOX:
This fixes 404 errors.
See https://github.com/zerodytrash/Simple-YouTube-Age-Restriction-Bypass/issues/18#issuecomment-864315540 for more information about this issue

Fixes #939

### DIFF
--- a/lib/info.js
+++ b/lib/info.js
@@ -325,6 +325,8 @@ const VIDEO_EURL = 'https://youtube.googleapis.com/v/';
 const getVideoInfoPage = async(id, options) => {
   const url = new URL(`https://${INFO_HOST}${INFO_PATH}`);
   url.searchParams.set('video_id', id);
+  url.searchParams.set('c', 'TVHTML5');
+  url.searchParams.set('cver', '7.20190319');
   url.searchParams.set('eurl', VIDEO_EURL + id);
   url.searchParams.set('ps', 'default');
   url.searchParams.set('gl', 'US');
--- a/lib/info.js
+++ b/lib/info.js
@@ -275,7 +275,7 @@ const getWatchJSONPage = async(id, options) => {
   let cookie = reqOptions.headers.Cookie || reqOptions.headers.cookie;
   reqOptions.headers = Object.assign({
     'x-youtube-client-name': '1',
-    'x-youtube-client-version': '2.20201203.06.00',
+    'x-youtube-client-version': cver,
     'x-youtube-identity-token': exports.cookieCache.get(cookie || 'browser') || '',
   }, reqOptions.headers);
 
--- a/lib/info.js
+++ b/lib/info.js
@@ -18,6 +18,8 @@ const BASE_URL = 'https://www.youtube.com/watch?v=';
 exports.cache = new Cache();
 exports.cookieCache = new Cache(1000 * 60 * 60 * 24);
 exports.watchPageCache = new Cache();
+// Cache for cver used in getVideoInfoPage
+let cver = '2.20210622.10.00';
 
 
 // Special error class used to determine if an error is unrecoverable,
@@ -307,6 +309,7 @@ const getWatchHTMLPage = async(id, options) => {
   let body = await getWatchHTMLPageBody(id, options);
   let info = { page: 'watch' };
   try {
+    cver = utils.between(body, '{"key":"cver","value":"', '"}');
     info.player_response = findJSON('watch.html', 'player_response',
       body, /\bytInitialPlayerResponse\s*=\s*\{/i, '\n', '{');
   } catch (err) {
@@ -326,7 +329,7 @@ const getVideoInfoPage = async(id, options) => {
   const url = new URL(`https://${INFO_HOST}${INFO_PATH}`);
   url.searchParams.set('video_id', id);
   url.searchParams.set('c', 'TVHTML5');
-  url.searchParams.set('cver', '7.20190319');
+  url.searchParams.set('cver', `7${cver.substr(1)}`);
   url.searchParams.set('eurl', VIDEO_EURL + id);
   url.searchParams.set('ps', 'default');
   url.searchParams.set('gl', 'US');
--- a/lib/info-extras.js
+++ b/lib/info-extras.js
@@ -282,8 +282,9 @@ exports.cleanVideoDetails = (videoDetails, info) => {
 
   // Use more reliable `lengthSeconds` from `playerMicroformatRenderer`.
   videoDetails.lengthSeconds =
-    info.player_response.microformat &&
-    info.player_response.microformat.playerMicroformatRenderer.lengthSeconds || info.player_response.videoDetails.lengthSeconds;
+    (info.player_response.microformat &&
+    info.player_response.microformat.playerMicroformatRenderer.lengthSeconds) ||
+    info.player_response.videoDetails.lengthSeconds;
   return videoDetails;
 };
 
--- a/lib/info-extras.js
+++ b/lib/info-extras.js
@@ -283,7 +283,7 @@ exports.cleanVideoDetails = (videoDetails, info) => {
   // Use more reliable `lengthSeconds` from `playerMicroformatRenderer`.
   videoDetails.lengthSeconds =
     info.player_response.microformat &&
-    info.player_response.microformat.playerMicroformatRenderer.lengthSeconds || info.player_response.lengthSeconds;
+    info.player_response.microformat.playerMicroformatRenderer.lengthSeconds || info.player_response.videoDetails.lengthSeconds;
   return videoDetails;
 };
 
--- a/lib/info-extras.js
+++ b/lib/info-extras.js
@@ -283,7 +283,7 @@ exports.cleanVideoDetails = (videoDetails, info) => {
   // Use more reliable `lengthSeconds` from `playerMicroformatRenderer`.
   videoDetails.lengthSeconds =
     info.player_response.microformat &&
-    info.player_response.microformat.playerMicroformatRenderer.lengthSeconds;
+    info.player_response.microformat.playerMicroformatRenderer.lengthSeconds || info.player_response.lengthSeconds;
   return videoDetails;
 };
 


AlbinGrans:
I don't have get any 404 error anymore with this fix, but seems like a side effect is that `videoDetails.publishDate` and `videoDetails.lengthSeconds` is undefined most of the time

Jaques22:
this fixes the 404 error but it also seems to add about 5 second delay to a getInfo as well

lib/info.js
```diff
@@ -325,6 +325,8 @@ const VIDEO_EURL = 'https://youtube.googleapis.com/v/';
 const getVideoInfoPage = async(id, options) => {
   const url = new URL(`https://${INFO_HOST}${INFO_PATH}`);
   url.searchParams.set('video_id', id);
+  url.searchParams.set('c', 'TVHTML5');
+  url.searchParams.set('cver', '7.20190319');
```
TimeForANinja, line 329: is it possible to get this number from one of the prev. fetched pages? not a big fan of hard coding version numbers... 😅

D3SOX: Not sure, I got the information from @zerodytrash. StackOverflow answers also suggest `6.20180913` for the version
https://stackoverflow.com/questions/67615278/get-video-info-youtube-endpoint-suddenly-returning-404-not-found

I tried it with other random numbers and strings and it seems to work too so I guess it only has to be set to something.

D3SOX:
@AlbinGrans Are those defined when you send the request without those parameters? Or is it just giving a 404 instead?
Maybe they renamed some parameters. I haven't had a look into those.

Jaques22:
> > seems to add about 5 second delay
> 
> @Jaques22 Hm I didn't notice any delay. Are you sure this is the culprit?

I should note that I am using ytdl-core-discord which relies upon this so maybe a problem with that wrapper.
I just did some very basic testing and it seems like the getinfo call takes longer now
`try {
                
                const results = await youtube.searchVideos(search, 1, { part: "snippet" });
                console.log(results);
                console.log("start getInfo");
                songInfo = await ytdl.getInfo(results[0].url);
                console.log("end getInfo");
                
                `
results definitely returns before the "start" log then it's about 5 seconds before the "end" log whereas before it was maybe 1-2s.
Though with this being a youtube api change its probably on the api side not ytdl

sudo-carson:
New endpoint removes `playerMicroformatRenderer` metadata as well

TimeForANinja:
> results definitely returns before the "start" log then it's about 5 seconds before the "end" log whereas before it was maybe 1-2s.
Though with this being a youtube api change its probably on the api side not ytdl

have heard of that befor in another context 🤔 so this might not be related to the 404's

Jaques22:
> > results definitely returns before the "start" log then it's about 5 seconds before the "end" log whereas before it was maybe 1-2s.
> > Though with this being a youtube api change its probably on the api side not ytdl
> 
> have heard of that befor in another context 🤔 so this might not be related to the 404's

Yeh it's probably not related, just thought I'd add in case a faster alternative was doable but I guess we'll see if someone comes up with one

lib/info.js
TimeForANinja: actually just searching the watch page gave me this
image
`cver: 7.20190319` is over 1 year old...

D3SOX: Should I change it to that value?

D3SOX:
The metadata seems to have changed with this. The duration no longer works. I might investigate this later today. If anyone has an idea let me know

JeHwanYoo:
it perfectly works thank you

lib/info.js
amitojsingh366: setting `cver` to `2.20210621.02.00` seems to cause the issue all over again

amitojsingh366: ok seems like to set `cver` to `2.20210621.02.00`,you need to set `c` to `WEB`. but when i did this it still caused 404s sometimes

amitojsingh366: for now i am applying a local patch with `c` set to `TVHTML5` and `cver` set to `6.20180913`

Zigecek:
Hi, when will be this merged?

D3SOX:
@timbly5000 I personally use [patch-package](https://github.com/ds300/patch-package) for now. You can apply the patch inside `node_modules` and then run `patch-package node-ytdl-core` or put this into a file named `ytdl-core+4.8.2.patch` in a `patches` folder relative the root of your project
```diff
diff --git a/node_modules/ytdl-core/lib/info.js b/node_modules/ytdl-core/lib/info.js
index 417f876..a2dd09a 100644
--- a/node_modules/ytdl-core/lib/info.js
+++ b/node_modules/ytdl-core/lib/info.js
@@ -325,6 +325,8 @@ const VIDEO_EURL = 'https://youtube.googleapis.com/v/';
 const getVideoInfoPage = async(id, options) => {
   const url = new URL(`https://${INFO_HOST}${INFO_PATH}`);
   url.searchParams.set('video_id', id);
+  url.searchParams.set('c', 'TVHTML5');
+  url.searchParams.set('cver', '7.20190319');
   url.searchParams.set('eurl', VIDEO_EURL + id);
   url.searchParams.set('ps', 'default');
   url.searchParams.set('gl', 'US');
```

D3SOX:
Only `player_response.videoDetails.lengthSeconds` seems to be set with these parameters. I added it a fallback to be included in the info result's `videoDetails` property

@AlbinGrans Can you confirm that it's working now?

AlbinGrans:
> Only `player_response.videoDetails.lengthSeconds` seems to be set with these parameters. I added it as a fallback to be included in the info result's `videoDetails` property
> 
> @AlbinGrans Can you confirm that it's working now?

Yep `lengthSeconds` is there now! `publishDate` is still undefined.

1001Josias (approved):
This change worked perfectly for me.

yalanyali (approved):

TimeForANinja:
ok - thinking about giving this a pass, even though this breaks basically all videoDetails
it's [the fix youtube-dl went for](https://github.com/ytdl-org/youtube-dl/issues/29333) and i don't see any other option
did test some with other `cver` but no success... hope this doesn't end in a "update cver every 3 months"

skick1234:
> did test some with other `cver` but no success... hope this doesn't end in a "update cver every 3 months"
We can get the client version from the web and replace `2.version` with `7.version` to change it to TV client version.
Like  `2.20210622.10.00` -> `7.20210622.10.00`

skick1234:
> did test some with other `cver` but no success... hope this doesn't end in a "update cver every 3 months"
We can get the client version from the web and replace `2.version` with `7.version` to change it to TV client version.

Like  `2.20210622.10.00` -> `7.20210622.10.00`

mpirescarvalho:
It's not fully working for me, still getting 404 in, let's say, half of the requests

TimeForANinja:
> > did test some with other `cver` but no success... hope this doesn't end in a "update cver every 3 months"
> 
> We can get the client version from the web and replace `2.version` with `7.version` to change it to TV client version.
> 
> Like `2.20210622.10.00` -> `7.20210622.10.00`

that sound's promising... didn't try that yet

D3SOX:
> > did test some with other `cver` but no success... hope this doesn't end in a "update cver every 3 months"
> 
> We can get the client version from the web and replace `2.version` with `7.version` to change it to TV client version.
> 
> Like `2.20210622.10.00` -> `7.20210622.10.00`

Are you sure the version numbers work like that? If so, we would need to pass the information from `getWatchHTMLPage`. 

I did a little test (with #946 applied) and this seems to do it:

```js
const clientVersion = info.response.responseContext.serviceTrackingParams
    .find(param => param.service === 'CSI').params
    .find(param => param.key === 'cver').value;
const tvVersion = `7${clientVersion.substring(1)}`;
```
(quick and dirty without any sanity checks)

If this is going to be the way to go, how do we want to implement it? I thought about setting it to something like `options.tvVersion` and then using it inside `getVideoInfoPage`
And what should we use as a fallback value if `getWatchHTMLPage` fails?

TimeForANinja:
thought about using `utils.between(body, '{"key":"cver","value":"', '"}')`
that would make it independant of finding a player config

besides that, i don't think that
> What should we use as a fallback value if `getWatchHTMLPage` fails?

is a real problem. If `getWatchHTMLPage` completely fails something is wrong with the video. If we don't find a player config we should still find the `cver`

another option would be just caching it in a global variable... default value could be sth. like `7.20210622.10.00`, update it every time we do any `getWatchHTMLPage` and it would therefor always be present

jbelford:
Thought I'd share this PR on lavaplayer https://github.com/sedmelluq/lavaplayer/pull/640

Tested on my own branch and `c=ANDROID` and `cver=16.24` also works

mome0320:
> Thought I'd share this PR on lavaplayer [sedmelluq/lavaplayer#640](https://github.com/sedmelluq/lavaplayer/pull/640)
> 
> Tested on my own branch and `c=ANDROID` and `cver=16.24` also works

It also `lengthSeconds`, `publishDate` is still undefined.

D3SOX:
@TimeForANinja Thank you, that should do it.

Lazy-C0der (commented):
at line 312 in info.js can you explain me what actually this code do ??

omenocal:
@TimeForANinja thanks for looking into this issue. I've got a large project that heavily relies on this library, do you have an ETA when a new release with these fixes will be published to NPM? Thanks,

TimeForANinja:
> @TimeForANinja thanks for looking into this issue. I've got a large project that heavily relies on this library, do you have an ETA when a new release with these fixes will be published to NPM? Thanks,

Waiting for some feedback from the other devs
Might merge after that

TimeForANinja:
Thanks again for the PR @D3SOX and everyone else who helped
For further Problems with (especially) video details check out https://github.com/fent/node-ytdl-core/issues/958

PR #972, opened 2021-07-16
add playerConfig types

MatteZ02:
This pr will add the typings for playerConfig under `videoInfo.player_response`.


TimeForANinja:
not a fan of adding typings for `playerConfig` since there's a high chance it's inconsistent but here we go
thanks for the pr

PR #984, opened 2021-08-01
Fixed URL decection

Histmy:
Hello, I noticed two bugs in URL detection and decided to fix them.

The first one was causing `validateURL` to be `true` with links like: `https://youtubedcom/v/SBjQ9tuuTJQ` or `http://www.youtube com/embed/SBjQ9tuuTJQ`. (Basically allowing any character between "youtube" and "com".)

And the second one made `getURLVideoID` return different id than YouTube would use in links like: `https://youtube.com/v/SBjQ9tuuTJQ/dQw4w9WgXcQ` or `https://youtube.com/SBjQ9tuuTJQ/somethink/somethink/dQw4w9WgXcQ`. (Basically YouTube uses the first string after `/v/` and ytdl was using the one after last `/`.)

I hope I made everything clear and actually helped instead of causing some problems. 😀

### DIFF
--- a/lib/url-utils.js
+++ b/lib/url-utils.js
@@ -28,7 +28,7 @@ exports.getURLVideoID = link => {
   let id = parsed.searchParams.get('v');
   if (validPathDomains.test(link) && !id) {
     const paths = parsed.pathname.split('/');
-    id = paths[2];
+    id = paths[1] === "v" ? paths[2] : paths[1];
   } else if (parsed.hostname && !validQueryDomains.has(parsed.hostname)) {
     throw Error('Not a YouTube domain');
   }
--- a/lib/url-utils.js
+++ b/lib/url-utils.js
@@ -28,7 +28,7 @@ exports.getURLVideoID = link => {
   let id = parsed.searchParams.get('v');
   if (validPathDomains.test(link) && !id) {
     const paths = parsed.pathname.split('/');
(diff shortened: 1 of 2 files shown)


redbrain:
Thanks for the PR! Good find on the first fix.
Second fix seems like an extreme edge case; is there a scenario in which this is needed?

Histmy:
Well, I cannot really think of any where it would break something, but it makes it behave like YouTube or Discord embeds would be handling it.

redbrain:
Well then, the PR looks fine to merge.

TimeForANinja:
would be cool if you could add test cases for these 2 scenarios so that we don't break it again in the future
should be pretty straight forward

else i can do it 😉

Histmy:
I think I added the tests, and hopefully correctly. I also fixed bug I introduced by the first commit. Check if everything is OK 😀

TimeForANinja:
your new version failed with links like `youtube.com/embed/VALID_ID`
switched to simply check for the hostname
e.g. `youtu.be/VALID_ID`
and `youtube.com/some word/VALID_ID/whatever`
tests are passing - linting is passing - code looks fine to me
would merge if fine with you?

Histmy:
Sure. I am glad I could at least help with this.

TimeForANinja:
nice 👍 
thanks for the pr 😉

PR #986, opened 2021-08-06
fix: `videoDetails.age_restricted` is always `false` in other languages

skick1234:
In other languages, notice url is not the value of `notice_url` property.

In `ar`:
```js
{
  'إشعار': 'هذا الفيديو محظور على فئات عمرية معيّنة (وفقًا لإرشادات المنتدى)',
  'إشعار_url': 'http://www.youtube.com/t/community_guidelines'
}
```
In `en-US`:
```js
{
  notice: 'Age-restricted video (based on Community Guidelines)',
  notice_url: 'http://www.youtube.com/t/community_guidelines'
}
```
In `vi-VN`:
```js
{
  'thông báo': 'Video bị giới hạn độ tuổi người xem (dựa trên Nguyên tắc cộng đồng)',
  'thông báo_url': 'http://www.youtube.com/t/community_guidelines'
}
```

I can check only keys have `_url` suffix but it is longer and YouTube may localize it like other keys in the future.


skick1234:
I will do it tomorow :>

PR #1022, opened 2021-10-14
Direct cipher signature & n-transform functions to circumvent throttling. 

### LINKED ISSUE #964: Randomly slow youtube download speed
Hello,

Important: The problem is random, maybe 1 chance on 8 to produce it. You have to download several videos in a row (about 10) to notice it.

Since a few weeks, randomly a youtube video can be slowed down to 48 Kio/s, so it takes 5-10 minutes to download a short video of 5 minutes instead of 4, 5 seconds, often the download does not succeed and stops after a few minutes.

This happens on several servers, several internet providers as well as with my private connection.

base on this: https://github.com/ytdl-org/youtube-dl/issues/29326

gatecrasher777:
Change sig.js to run base.js code directly for signature decipher and n-parameter transform.  Less code too. :)
Since Google's base.js is browser code it should not pose a security risk.
sig-test.js also changed - some of the tests were no longer useful with this methodology.

Fixes #964

Many projects are struggling with this throttling issue.


TimeForANinja:
> Since Google's base.js is browser code it should not pose a security risk.

🤔 bold claim - in the end we execute code straight from the internet and hope google is playing nice

LuanRT:
> FWIW I found I could could get full access to the all node-js modules from within vm running an empty context... just by using a this.constructor inside the sandbox. So require('fs') and delete all is quite possible. That's why vm2 is preferred. No known way to break out of the sandbox.
> 
> I would concede that a reliable means of n-transform without running any google code directly is first prize. But it might just be inherently unreliable by design, through the constant introduction of novel transformations. That seems to have been the experience so far of those trying to emulate it, although I can't be certain about that. Perhaps it will stabilize.

Any progress on this? If it helps, I have developed a [script](https://github.com/LuanRT/YouTube.js/blob/main/lib/NToken.js) that parses the JavaScript code and emulates the necessary transformations on the n token— it was not based on VLC's approach but I'm constantly debugging it. So far I have tested it with 20+ different players and only ran into minor parsing problems on one of them, however they were easily fixed. I guess one could port it to ytdl-core if executing the n transform function directly is a problem.
(do note that my library is also written in JavaScript, so it should be pretty simple to port).

Michal-Szczepaniak:
Works for me but didn't test thoroughly

peepopoggers:
:+1:

Michal-Szczepaniak:
Tested by 3 people, not only it improves performance of getting url by a lot, it fixed throttling completely, this patch is amazing can we merge it asap?

gazugafan:
Working for me so far. Going to get tested more thoroughly tonight.

KuraimiKun:
> 

how can i do that

TimeForANinja:
thanks again for the pr - should be released on npm in a second
i opened #1043 as a reminder / discussion to remove that vm-dependency
for now this looks like the best fix

Issue #1036, opened 2021-11-26
[SUPPORT] Handle YT-returns-wrong-length bug by manually setting stream.end = true in end event.


mepc36:
**N.B.: THIS ISSUE IS NOT TO FIX THIS PACKAGE; IT IS TO OFFER PRODUCT SUPPORT TO USERS OF THIS PACKAGE. THE ACTUAL FIX IS TO CHANGE THE HEADERS ON YOUTUBE'S SIDE.**

### Before creating an issue, please try the following

- Are you on the latest version of ytdl-core?

Yes

- Search the list of issues for a possible duplicate.

None found.

- If your question is about using the library for a specific use case, check out the [examples folder](https://github.com/fent/node-ytdl-core/tree/master/example).

### Otherwise, feel free to open an issue

Include the following if your issue is a bug

- If you're getting a specific error message, include the full stack trace

**These logs come from the youtube-mp3-downloader package, which wraps this one:**

mediafetcher_1  | $$$ FETCHING media for Juice WRLD - "Lost Cause"
mediafetcher_1  | $$$ youtubeVideoId for Juice WRLD - "Lost Cause": Kl6s21WoZwE
mediafetcher_1  | Juice WRLD - "Lost Cause": {"videoId":"Kl6s21WoZwE","progress":{"percentage":3.593749657273325,"transferred":376832,"length":10485761,"remaining":10108929,"eta":148,"runtime":5,"delta":376832,"speed":68514.90909090909}}
mediafetcher_1  | ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
mediafetcher_1  | Juice WRLD - "Lost Cause": {"videoId":"Kl6s21WoZwE","progress":{"percentage":7.18749931454665,"transferred":753664,"length":10485761,"remaining":9732097,"eta":136,"runtime":10,"delta":376832,"speed":71777.52380952382}}
mediafetcher_1  | ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
mediafetcher_1  | Juice WRLD - "Lost Cause": {"videoId":"Kl6s21WoZwE","progress":{"percentage":10.781248971819975,"transferred":1130496,"length":10485761,"remaining":9355265,"eta":132,"runtime":15,"delta":376832,"speed":70656}}
[...]

PR #1037, opened 2021-11-26
Handle YT-returns-wrong-length bug by manually setting `stream.end = true` in `end` event.

mepc36:
See issue [#1038](https://github.com/fent/node-ytdl-core/issues/1036) for more info.

TL;DR - YT returns incorrect `content-length` header for some video; this code change handles such errors (somewhat inelegantly); I don't expect maintainers to accept this change because YT is the source of the problem (not ytdl-core); it's here to support other users running into same problem instead.


mepc36:
Closing until I can investigate header-handling more.

Issue #1034, opened 2021-11-22
song_url missing on Media interface type.


Auxority:
![image](https://user-images.githubusercontent.com/35202343/142899518-713e8c05-f409-490c-80dd-4eda51cb405b.png)
The song_url property exists when youtube detects music in videos. But this property is missing from the Media interface in the types declaration file.
A simple fix would be to just add the song_url property to the Media interface in index.d.ts.
```
    interface Media {
      category: string;
      category_url: string;
      game?: string;
      game_url?: string;
      year?: number;
      song?: string;
      song_url?: string; // Just add this and the problem is solved.
      artist?: string;
      artist_url?: string;
      writers?: string;
      licensed_by?: string;
      thumbnails: thumbnail[];
    }
```

PR #1069, opened 2022-02-28
Support multiple media objects

### LINKED ISSUE #920: The list of media is not complete.
The media list seems to return only the last one.

Capture d’écran 2021-05-13 à 11 19 38

Capture d’écran 2021-05-13 à 05 31 08

Capture d’écran 2021-05-13 à 05 30 48

— ytdl-core version: `4.7.0`

chrisdeaner:
**NOTE: This could be a breaking change. The proposed type returned is changing from an object to an array.**

Currently the ytdl-core library supports returning a Media object as part of the Video Details. This approach has a couple of issues:

1) When there are multiple songs returned for a video, the Media object can contain information from both songs if the Media attributes are not exactly the same.

Example: Imagine a video contains two songs, Foo and Bar, returned in that order. The Foo song contains only info about the artist, "FooRealz" and the Bar song contains only info about the writer, "BarWritez". Currently the object returned would look like so, squashing the two songs into one:
```json
{
  "media": {
    "song": "Bar"
    "artist": "FooRealz",
    "writers": "BarWritez"
  }
}
```
A good example of this issue can be found here: https://www.youtube.com/watch?v=4bMM7tGV9MI where data from Ozzy Osborn and Ludacris get merged. 

2) Similarly, when videos contain more than one song, all expected information will not be returned.

This PR creates a media array to return all the information present, adds and updates tests, and updates type descriptions. 

Edit:
closes #920


TimeForANinja:
would be cool to also use this to close #1034

chrisdeaner:
@TimeForANinja - Would be great if this PR helps resolve outstanding issues!

A question about tests: `age_restriction` related tests are failing for me. Specifically "Returns correct video metainfo" and "Returns correct age_restricted property in other languages". I see that you released 4.11.0 yesterday with some `age_restriction` changes. Are these test failures expected or should I take these failures as an issue with my code / environment? 

Any help is appreciated.

TimeForANinja:
i added some messages to identify the failing assertions
unfortunately you broke [this](https://github.com/chrisdeaner/node-ytdl-core/blob/multiple-media-tags/lib/info.js#L88) part of the original code with your changes

guess it has to do with not pushing `media` into `medias` after every iteration of [this](https://github.com/chrisdeaner/node-ytdl-core/blob/multiple-media-tags/lib/info-extras.js#L38) for loop? though i must say i'm not quite sure what these 
````js
if (title === 'song') {
          media = {};
          medias.push(media);
        }
```
are supposed to achieve any way 😅

chrisdeaner:
@TimeForANinja Thanks for adding that messaging and for your comments; very helpful. I pushed some changes and all the tests are now passing. As for the the blocks of code with `title === 'song'`, they designate when the information for a song is starting. I could see no other way where the data indicates that it contained discrete song information. This approach works consistently in my tests. 

Looking forward to getting this in! Thanks for your help and review!

absidue:
@chrisdeaner As you are checking for the title in English this PR only works if the `lang` option is set to `en` or left blank (the default value of the `lang` option is `en`).

chrisdeaner:
@absidue that makes perfect sense. I'm going to close this PR for now and reassess the approach. I'm not very familiar with this library so if you can think of any viable approaches, feel free to drop them in this PR and I'll look into it.

PR #1010, opened 2021-09-20
Fix for flagged videos

Bullfrog098:
I'm pretty sure you only need to put it here but i'm not 100% sure, everything works for me after putting this here.

Issue #1009, opened 2021-09-20
Community flagged videos coming back as 410


Bullfrog098:
Trying to download a video that is flagged inappropriate or offensive (Ex. https://www.youtube.com/watch?v=hc0ZDaAZQT0) comes back as a 410. This was fixed in youtube-dl with "&bpctr=9999999999" so all that needs to be done is when the base_url and id are added together add "&bpctr=9999999999" right after.

redbrain:
Thanks for the information! Do you have time to draft a PR to close the issue?

Bullfrog098:
Not sure if it needs to be placed elsewhere but its in #1010

MIGI10:
Adding that to the url doesn't fix the 410 errors for me... I have tried that together with passing a cookie and an identity token header but the errors persist. I'm able to download age restricted videos but not those flagged as sensitive content (like the one linked above)

Bullfrog098:
> Adding that to the url doesn't fix the 410 errors for me... 
sorry for late response but did you add "&bpctr=9999999999" into the ytdls code itself or did you just add it into the url you were requesting?

TimeForANinja:
should be fixed by #1022

cchen028:
still experiencing the issue on certain videos after updating to 4.9.2

TimeForANinja:
same, but #1010 is not fixing it for me 🤔

FadelesSpade:
This issue is still arising for all age restricted videos on 4.10.0 and #1010 is not fixing it either.

codeerror90:
can't fix it for me

pixelblob:
Any updates on this?

aliazhar-id:
Nothing, I'm still getting this problem, sad

TimeForANinja:
> Nothing, I'm still getting this problem, sad

you can try the #1080 and tell me if that fixed it for you 🧐

AntonyOnScript:
same issue here

umarengr99:
i m using ytdl-core npm package and facing Region Restriction error: which throw a generic Status code 410. 
how i can fix this issue by changing the region.

PR #1080, opened 2022-03-26
Fix for watch page community restricted videos

JackReevies:
### Issue

Community age restricted videos on Youtube (the ones that require clicking the "I understand and wish to proceed" button before watching) were causing the module to throw a 410 when trying a `getInfo`

### Cause

`getInfo` runs through a maximum of 3 urls until it has enough metadata (as defined in the validate function on info.js:59). Often we can get enough data from the first url (the basic watch page) to satisfy this and exit early, unfortunately no formats are made available when a video requires the user to click that age restricted button. The other URLs are visited and the last one in the list (`youtube.com/get_video_info`) throws a 410 Gone, causing the module to also throw. 

### Solution

Appending `&bpctr=9999999999&has_verified=1` to the first url (the basic watch page) seems to bypass that "I understand and wish to proceed" button and makes the format/stream data available right away, thus eliminating the need to visit the other URLs and never calling out to the `youtube.com/get_video_info` url

### Related

I didn't discover this, I simply ran into the problem and was searching for solutions. There have been conversations on this repo in issues and PRs about this (https://github.com/fent/node-ytdl-core/issues/1009 and https://github.com/fent/node-ytdl-core/pull/1010) but they've been closed/dismissed as being fixed under https://github.com/fent/node-ytdl-core/pull/1022. This does not seem to be the case for me (and at least a few other users). 

Here's a video I've been using to test with https://www.youtube.com/watch?v=lwVhz0QCFEY (you can also see here that adding &bpctr=9999999999&has_verified=1 to the end of that URL gets rid of the prompt and starts playing immediately. 

Code can be tested with the following test function

```
async function start() {
	  try { 
	  	const info = await getInfo('lwVhz0QCFEY') 
	  } catch(e) {
	    console.error(e)
	  }
}
```

### DIFF
--- a/lib/info.js
+++ b/lib/info.js
@@ -126,7 +126,7 @@ const isNotYetBroadcasted = player_response => {
 };
 
 
-const getWatchHTMLURL = (id, options) => `${BASE_URL + id}&hl=${options.lang || 'en'}`;
+const getWatchHTMLURL = (id, options) => `${BASE_URL + id}&hl=${options.lang || 'en'}&bpctr=9999999999&has_verified=1`;
 const getWatchHTMLPageBody = (id, options) => {
   const url = getWatchHTMLURL(id, options);
   return exports.watchPageCache.getOrSet(url, () => utils.exposedMiniget(url, options).text());


bmckalip:
This works for me. I might suggest using `&bpctr=${Date.now()}&has_verified=1` instead of `&bpctr=9999999999&has_verified=1` as after inspecting the query params of the url from youtube's site, this seems to be more accurate. Both work for me, however.  

There might be a better way to do this overall, though. I'm just diving into the code of this project after running into the issue in my [app](https://github.com/bmckalip/zotero-media-downloader), and it seems like there is no way to specify arbitrary query params for the request made in `getWatchHTMLURL`. I may try my hand at  creating something more generic.

pixelblob (approved):

pixelblob:
when can this be merged?

PR #1084, opened 2022-04-11
New Feature: getVideoPlaybackURL function

twarped:
adds two functions, one is getFormat, and the other is getVideoPlaybackURL. getformat just returns the formatutils.chooseformat without having to do info.formats. getVideoPlaybackURL gets the video playback url, so that you don't have to manually get it, and it's just easier. it doesn't mess with any of the code, and it's a simple addition that might make things easier.


TimeForANinja:
sorry, but i don't see the advantage of adding these two functions

PR #1187, opened 2023-03-06
Explicitly include range parameter in chunked stream URLs

### LINKED ISSUE #1186: Audio .mp3 download very slow when using ffmpeng
Example my code

```javascript
const readline = require('readline');
const ytdl = require('..');
const ffmpeg = require('fluent-ffmpeg');

let id = '7wNb0pHyGuI';

let stream = ytdl(id, {
  quality: 'highestaudio',
});

let start = Date.now();
ffmpeg(stream)
  .audioBitrate(128)
  .save(`${__dirname}/${id}.mp3`)
  .on('progress', p => {
    readline.cursorTo(process.stdout, 0);
    process.stdout.write(`${p.targetSize}kb downloaded`);
  })
  .on('end', () => {
    console.log(`\ndone, thanks - ${(Date.now() - start) / 1000}s`);
  });
```

Brainicism:
First time touching this repo, so not sure if I missed any edge cases, but creating this PR to start the conversation on the permanent fix. Credits to @gatecrasher777 for the idea.

Attempts to resolve #1186, where youtube is throttling requests without the 'range' parameter in the URL. 

Instead of setting `end` to 0 (falsey) as a condition to check for whether to continue chunking, we check whether `end` has reached the end of the file.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -143,19 +143,20 @@ const downloadFromInfoCallback = (stream, info, options) => {
         parseInt(format.contentLength);
 
       const getNextChunk = () => {
-        if (!rangeEnd && end >= contentLength) end = 0;
+        if (!rangeEnd && end >= contentLength) end = contentLength;
         if (rangeEnd && end > rangeEnd) end = rangeEnd;
-        shouldEnd = !end || end === rangeEnd;
+        shouldEnd = end === rangeEnd || end === contentLength;
 
         requestOptions.headers = Object.assign({}, requestOptions.headers, {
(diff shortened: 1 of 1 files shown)


ItsPerona (approved):

Yonle (changes requested):
Usually, If your end range is the same as the full `Content-Length`, You will ended up getting 416 error.

To fix this, You may `contentLength-1`

Yonle:
The `&range` parameter did not seems to correctly working at some cases. Like `1-10`.

PR #1188, opened 2023-03-09
fix: preserve typescript autocomplete

Mist3rBru:
When we type something like `'literal-string' | string` we loose all the typescript autocomplete. My proposal is to maintain either typescript autocomplete and generic strings by using the `Omit` hack.

```ts
// before
type TString = 'foo' | 'bar' | string

const test: TString = 'any' // no autocomplete and no error
const test2: TString = 'foo' // no autocomplete and no error


// after 
type TString = 'foo' | 'bar' | Omit<string, 'foo' | 'bar'>

const test: TString = 'any' // no autocomplete and no error
const test2: TString = 'foo' // with autocomplete and no error
```

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1,7 +1,6 @@
 declare module 'ytdl-core' {
-  import { ClientRequest } from 'http';
   import { Readable } from 'stream';
-
+  
   namespace ytdl {
     type Filter = 'audioandvideo' | 'videoandaudio' | 'video' | 'videoonly' | 'audio' | 'audioonly' | ((format: videoFormat) => boolean);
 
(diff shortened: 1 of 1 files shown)


typings/index.d.ts
```diff
@@ -11,8 +10,10 @@ declare module 'ytdl-core' {
       requestOptions?: {};
     }
 
+    type ChooseFormatQuality = 'lowest' | 'highest' | 'highestaudio' | 'lowestaudio' | 'highestvideo' | 'lowestvideo'
+
     interface chooseFormatOptions {
-      quality?: 'lowest' | 'highest' | 'highestaudio' | 'lowestaudio' | 'highestvideo' | 'lowestvideo' | string | number | string[] | number[];
+      quality?: ChooseFormatQuality | Omit<string, ChooseFormatQuality> | number | (ChooseFormatQuality | Omit<string, ChooseFormatQuality>)[] | number[];
```
TimeForANinja, line 16: Can be make this a bit more beautiful with sth like?
```ts
type SingleOrArray<T> = T | T[];
```

other than that, although i think this is more of a interpreter/linter bug, i don't see any harm in merging this

Mist3rBru: I made it thought:
```ts
type ExtendString<T extends string> = T | Omit<string, T>
...
interface chooseFormatOptions {
      quality?: ExtendString<ChooseFormatQuality> | number | ExtendString<ChooseFormatQuality>[] | number[]
...
```
What do you think? Maybe the name could be better, but that was the best that I thought

TimeForANinja: i didn't even think of that 👍

PR #1202, opened 2023-04-06
Fix: Missing catch or finally after try

oskar2594:
To resolve the problem of the missing catch block in the YouTube Script, I added some simple code to include the missing catch block in the relevant function. I have tested the built-in tests and did not encounter any errors. If you would like to test the functionality yourself, please update the player script for testing purposes.


Sansekai:
Still error for me

```
"ytdl-core": "github:oskar2594/node-ytdl-core"
```

